PR #18, opened 2019-01-08
Move package-relevant files to new directory

midgleyc:
I've moved the node-relevant files into a 'src' directory. This should stop various files and folders (azure-pipelines.yml, .git) getting packaged into the final deployment.

Tests, if created, should also go outside this directory.


AdamChrimes:
Superseded by https://github.com/nhsuk/nhsuk-service-manual/pull/21

PR #27, opened 2019-01-15
Design example

mikemonteith:
Design example code taken from a mixture of torchbox's component (https://nhs-service-manual-alpha.staging.torchbox.com/service-manual/patterns/header/) and the govuk component (https://design-system.service.gov.uk/components/breadcrumbs/).

I've only implemented one code example on the breadcrumb page. After this MR is accepted we should be able to add more.

Other changes:
- Javascript is now compiled with rollup and babel as an extra gulp task. `npx gulp javascript`. 
- `/design-example/<example>` URLs will render an example snippet in a standalone html template.


mikemonteith:
To add a code example to a page:

```
{% from 'includes/design-example.njk' import designExample %}

{{ designExample({
    type: 'breadcrumb'
}) }}
```

`type` should be the type of example you want to display, and the example goes in `app/examples/<type>.njk`

mikemonteith:
At the moment, all examples will showr HTML and Nunjucks code with a standalone preview box. We can add other options (e.g don't show nunjucks) if we have any need for it.

PR #84, opened 2019-01-31
Fixed header and hero display issue on ie8

davidhunter08:
Fixed header and hero display issue on ie8


mcheung-nhs:
Has been fixed in the frontend library so no need to fix it here. Just need to update the library to the newest version.

PR #87, opened 2019-01-31
Tidy up OGL content and styling in the footer

mcheung-nhs:


AdamChrimes:
Looks good - should we use colour and breakpoint variables incase they ever change in the NHS.UK frontend (unlikely) to avoid having to manually change them in the service manual?

`$color_nhsuk-grey-4`
`$color_nhsuk-grey-1`

```
@include mq($from: tablet) {
    
  }
```

The rest of the `scss` file needs the same doing - I think its on the backlog as a separate ticket.

mcheung-nhs:
@AdamChrimes - agreed. Have used the frontend variables and mixins.

PR #110, opened 2019-02-05
Content amendments

sarawilcox:
Delete old patterns pages, now replaced by UI styles, components and patterns pages. Also change content in 4 new components files.


AdamChrimes:
Content review and tested on Chrome/Safari

PR #116, opened 2019-02-11
Move page sections into their subfolders

AdamChrimes:
Move page sections into their relevant subfolders and rename the pages to /index as they sections not individual pages. This fixes the issue if users have a trailing slash on the URL.

Example: `/service-manual/design-principles/` was broken because of the trailing slash, which is expected behavior as its looking for a page within the `design-principles` directory rather than the `design-princples.njk` page. 

Now you can visit `/service-manual/design-principles/` or `/service-manual/design-principles`

Added a temporary redirect for the content-style-guide page as it has been renamed.

Going to monitor the Azure traffic logs for anymore 404 errors but we probably should build in some logging at some point.


app.js
```diff
@@ -70,6 +70,10 @@ app.get('/', (req, res) => {
   res.redirect('/service-manual');
 });
 
+app.get('/service-manual/content-style-guide', (req, res) => {
```
mikemonteith: Might be a good idea to add a comment here explaining why we're doing this redirect.

mikemonteith:
What is the canonical URL for our pages? Ending with a slash or without? We should put in a redirect rule to always append (or remove) the `/`.

AdamChrimes:
@mikemonteith commented the redirect.

Whether there should be a '/' hasn't really been decided and is up for discussion, this was more of a quick fix before putting in some logic to add/remove the '/'/

Issue #117, opened 2019-02-11
404 page being served as a 200


mikemonteith:
![image](https://user-images.githubusercontent.com/227467/52556266-fef7f580-2de3-11e9-89a4-b8ddbf1bae1f.png)

The "page not found" page is being served as a valid http response (200 OK). It should be returning a 404 http code.

AdamChrimes:
Fixed in #120

PR #120, opened 2019-02-11
[refs #117] Set status code 404 on page not found pages

AdamChrimes:
The "page not found" page is being served as a valid http response (200 OK). It now returns a 404 status code.


mcheung-nhs:
404 returned:

Issue #126, opened 2019-02-20
Broken home link in the breadcrumb on the error message page


davidhunter08:
The home link in the breadcrumb on the [error page](https://beta.nhs.uk/service-manual/styles-components-patterns/error-message) is broken.

The URL is /service-manualindex

I did a search in the codebase and this was the only instance of it I could find.

Also, it might be better to create the breadcrumb for styles, component and pattern pages as a Nunjucks include to reduce code repetition?

PR #131, opened 2019-02-26
[refs #126] Fix broken breadcrumb home link on error message page

AdamChrimes:


app/views/styles-components-patterns/error-message.njk
```diff
@@ -11,7 +11,7 @@
 {{ breadcrumb({
   items: [
     {
-      href: "/service-manualindex",
+      href: "/service-manual/index",
```
mikemonteith: Shouldn't this just be `href: "/service-manual/`?

AdamChrimes: It should, fixed!

PR #135, opened 2019-02-27
Target build through PR same as to branch.

midgleyc:
Don't run both builds on PR, only the relevant one (and only if the PR targets a relevant branch).


midgleyc:
The non-terraform build is currently a required status check, so PRs targeting terraform only couldn't be merged with this change.

PR #166, opened 2019-03-12
Update standards files

sarawilcox:
Combine old points 2 and 3.
Delete old point 3.
Update index, about and points 1 and 2.


sarawilcox:
@ericlamarca, could you review this please? @AdamChrimes, for info too. Sorry - I should requested merging this into develop, *not into the master*. How do we change that?

PR #168, opened 2019-03-18
Improve the documentation for the layout page

AdamChrimes:


ericlamarca:
Hey @sarawilcox @AdamChrimes - page looks good to me. Only thing I'd wonder is what's the rationale for changing from grid to layout?

AdamChrimes:
Its a little more than grids, literally speaking grids would just be the column widths and nothing else. Layout contains the page wrappers, page containers, page widths, breakpoints, grid and potentially could include spacing.

Its also the standard term used for this, such as on [Bootstrap](https://getbootstrap.com/docs/4.3/layout/) and on the [GOV.UK design system](https://design-system.service.gov.uk/styles/layout/)

ericlamarca:
Sure, that makes sense. Was asking because 'Grid' is sometimes used to cover those wider elements (in [BBC GEL](https://www.bbc.co.uk/gel/guidelines/grid) or [Photon](https://design.firefox.com/photon/visuals/grid.html) for example). And guess we haven't seen an issue with this so far? 

But I do see the point, and it makes sense to align with GOV on this. Just checked @sarawilcox also agrees to go ahead 👍

PR #169, opened 2019-03-19
Cookie consent

bradmorton1:
this change will update the cookie consent solution to be in line with the rest of beta.nhs.uk. currently WIP whilst the rest of beta is implemented. Cookie policy pages will be hosted outside of the service manual


app/views/includes/layout.njk
```diff
@@ -53,7 +56,7 @@
 
   <body>
     <!-- Google Tag Manager (noscript) -->
-    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-W8BPX7Q"
+    <noscript><iframe data-src="https://example.com" data-cookieconsent="statistics" src="https://www.googletagmanager.com/ns.html?id=GTM-W8BPX7Q"
```
mcheung-nhs: @bradmorton1 - what does the `data-src` attribute do and should it be `https://example.com`?

bradmorton1: For iframes, you need to use data-src instead of the type attribute in scripts to be able to control whether it is shown or not. Fixed it as it wasn't quite right when I ported it across.

The cookie code should read the data source, then set the source of the iframe if consent has been accepted

mikemonteith: Why do we even have google tag manager in the service manual anyway? We're not running ads.

bradmorton1:
This is currently blocked as service manual has been descoped from cookie changes

PR #178, opened 2019-04-10
Patterns

bencullimore:
here's my write up for a-z and mini hub


AdamChrimes:
The build is failing because there needs to be an empty new line at the end of the `main.scss` file.

```
app/styles/main.scss
  8:40  error  Files must end with a new line  final-newline
```

app/styles/main.scss
```diff
@@ -5,4 +5,4 @@
 @import 'app/main';
 
 // Design examples
-@import 'design-example/design-example';
+@import 'design-example/design-example';
```
AdamChrimes: Here is where the empty new line is needed.

bencullimore:
Hey @ericlamarca I'm not too sure about introducing another term for essentially the same thing - we had lots of confusion at the beginning with patterns and components, then we had that naming session. I'd be wary of introducing another term.

I also share your worry about duplication... but that's based on the assumption that users know all the components that are in the manual. My hypothesis is that users might visit the patterns more than the components - because they're full things. Something to test with users?

ericlamarca:
Hey @bencullimore - yes, exactly. So, do we need to change the h2s 'Elements of the mini-hub' and 'Elements of the A-Z' to say components instead?

Right, I see what you mean, by all means let's test it with users. In the meantime, so all components published in the manual are treated consistently, should we add these new ones to the components list as well?

AdamChrimes:
Merging this to develop to review it on the staging environment.

Discussions above will continue before merging to master.

PR #182, opened 2019-04-23
Fixed hyphens in class names on typography page

anlinhs:


AdamChrimes:
Good spot, thanks @anlinhs 

I'll merge this to the staging environment before releasing to live.

PR #198, opened 2019-06-25
Revert numbers and measurements changes

AdamChrimes:
Revert the changes to the numbers and measurement pages and the navigation across other sections until they have been completed.

These changes have been stored on the `feature/numbers-and-measurements` branch to be picked it up at a later date.


AdamChrimes:
@sarawilcox would you just be able to confirm whether this covers all the places? 

Here's the original PR https://github.com/nhsuk/nhsuk-service-manual/pull/194

PR #272, opened 2019-08-06
Update A to Z with entries for health record and medical record

sarawilcox:
2 new terms in the content style A to Z after 18 July Style Council meeting.


AdamChrimes:
Hi @sarawilcox please could you add a change log entry into the `CHANGELOG.md` file

sarawilcox:
I've updated the change log, @AdamChrimes . A few more entries to the A to Z coming today in the GitHub training.

AdamChrimes:
Thanks @sarawilcox, if we could encourage them to update the CHANGELOG in those sessions that would be great.

PR #275, opened 2019-08-07
Carecardchanges

sarawilcox:
Update card card page with new content for content designers, as agreed at 18 July Style Council meeting 

@AdamChrimes or @davidhunter08, if you could merge this with develop please, I will ask a few content designers to check it on staging before we publish. Thanks a lot.


sarawilcox:
BTW, could you please check that I've applied the care card code correctly in the furosemide example under the Care card content heading? I copied in the html from higher up the page, so I'm assuming it's OK. Thanks a lot.

AdamChrimes:
Looks good @sarawilcox

PR #280, opened 2019-08-07
Amend apostrophes so they're straight

SusieWd:
Please. Old apostrophes were curly, we need straight ones.


sarawilcox:
Thanks @SusieWd. I'll review this tomorrow.

PR #285, opened 2019-08-08
Accessibility: updated header home link aria label

davidhunter08:
The home link in the header was labelled as 'NHS homepage', causing confusion for users using screenreaders.

The aria label has been updated to 'NHS Digital service manual homepage'


sarawilcox:
@davidhunter08 ideally we should make the d in digital lower case but maybe not a big issue if the aria label will only be used for voice?

AdamChrimes:
I've made the `D` lowercase, for consistency.

PR #328, opened 2019-09-05
Limit search dropdown results to 10

sabi98:


AdamChrimes:
I've rebased this with master to fix the issue with the broken link on the sitemap page.

PR #354, opened 2019-10-01
Remove wizard behind the curtain example

JackMatthams:
Remove wizard behind the curtain example following user research feedback


sarawilcox:
Brilliant, thanks very much Jack.

PR #361, opened 2019-10-01
Amend nav and links

sarawilcox:
Change order of pages
Rename some pages
Update nav and links, including to new sensitivities page

### DIFF
--- a/app/views/content/how-to-write-good-questions/1-understand-the-problem-before-you-write-your-questions.njk
+++ b/app/views/content/how-to-write-good-questions/1-understand-the-problem-before-you-write-your-questions.njk
@@ -50,7 +50,7 @@
     </ul>
     <p>It will help if your team also</p>
     <ul>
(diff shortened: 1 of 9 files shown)


cjforms:
Hi Sara, 

1. Is there a way to review these changes other than reading the compared files? 

2. Please can you say a bit more about why changes are happening, such as whether the changes have come from user research (and with whom) or from expert review (and with whom?)

thanks

sarawilcox:
Hi @cjforms, I'm afraid comparing the files is the only way for now. 

@davidhunter08 and I are making a lot of changes at the moment based on feedback we've had and the design review. Most of the changes are about moving things around, streamlining them and making them easier to understand and navigate. We're then going into a final round of testing early next week, setting people some info-finding tasks as well as seeing how they navigate the whole (dip and lure/dip and learn?). 

It would be good if we could arrange a chat with you to fill you in before we go live - within the next few weeks. I'll get in touch towards the end of this sprint (next Thursday) when we'll be clearer of our launch date. Hope that's OK.

cjforms:
OK, thanks. I'll ignore changes in this area for a bit then and look forward to our chat and the launch date.

PR #379, opened 2019-10-15
Update single use of 'Dos' and 'Don'ts' guidance

davidhunter08:


davidhunter08:
> Thanks for doing this. There is a typo. It should read: if there aren't any Don'ts.

That's been fixed.

PR #424, opened 2019-11-06
Fix summary list css class documentation

mikemonteith:
Use the correct `--no-border` css modifier, not `--without-border`


davidhunter08:
Thank you @mikemonteith

PR #432, opened 2019-11-11
HTML cleanup

davidhunter08:
Tidy HTML for all pages.

- Consistent spacing and indenting to make the code easier to read
- Remove unnecessary spacing override classes
- Create 2 new page layouts (full width and two thirds) to reduce code duplication
- Add Sass styles to constrain widths of elements when using the full width layout


AdamChrimes:
Looks good to me

PR #446, opened 2019-11-22
Update nav in content style guide

sarawilcox:
Add in link to How to write good questions for forms.

### DIFF
--- a/app/views/content/partials/related-nav.njk
+++ b/app/views/content/partials/related-nav.njk
@@ -27,6 +27,11 @@
         <a href="/service-manual/content/health-literacy">Health literacy</a>
       </li>
       {% endif %}
+       {% if pageTitle != 'How to write good questions for forms' %}
+      <li class="nhsuk-related-nav__item">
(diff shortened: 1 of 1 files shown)


sarawilcox:
@AdamChrimes I've sorted out the link to the forms guidance but notice that the link to the Numbers, measurement page appears on the Numbers page, which it shouldn't do. Can you help please?

AdamChrimes:
Hi @sarawilcox we don't have a navigation on the 'How to write good questions for forms' pages, I don't know if this was a conscious decision with the design of this page but this change you have made won't have any affect as the related navigation isn't included on the page to begin with.

Code for including the related navigation:

```
{% block asideContent %}
  {% include './partials/related-nav.njk' %}
{% endblock %}
```

I've updated the 'Numbers, measurements, dates and time' page to remove the current active page.

sarawilcox:
Thanks for sorting the number page, Adam. 

Yes, we decided it didn't fit well with the current forms guidance page, because of the competing nav.

PR #462, opened 2019-12-13
Update inclusive language page

sarawilcox:
Make further changes based on recent feedback and review.


sarawilcox:
Hi @JackMatthams, I've updated this page but we need to go through some final queries with Jean on Monday.

PR #467, opened 2019-12-16
Parent page in search result

tomdoughty:
Gets the parent page from the breadcrumb and adds it to search results.
Works for results page or autocomplete results


middleware/page-index.js
```diff
@@ -71,60 +71,51 @@ class PageIndex {
...
+        h2,
+        h3,
+        parent,
+        title,
+        url,
+      },
+    ];
   }
 
   indexPageSpecial($, url, description) {
```
AdamChrimes: Can we remove this `indexPageSpecial` now that we no longer include `A to Z of NHS health writing` in the result title as it is in the result category?

davidhunter08:
The 'Introduction' page in 'How to write good questions' shows up as 'Content style guide' when it should say 'How to write good questions for forms'.

The other pages within  'How to write good questions for forms' do say 'How to write good questions for forms'.

Screenshot 2019-12-17 at 10 57 38
Screenshot 2019-12-17 at 10 57 18

middleware/page-index.js
tomdoughty: Not easily enough to make it worth while in my opinion... it has to do a fair bit of its own stuff to get the info from each H3 section.

AdamChrimes: Cool, lets not worry about it. We can always refactor in the future.

AdamChrimes:
Hey @tomdoughty we just need to rearrange some content on the 'Introduction page' for How to write good questions so that the parent page for it is correct and then we can merge this in.

https://jira.service.nhs.uk/browse/STAN-1546

sarawilcox:
@AdamChrimes We've rearranged the content in the redesign/nav branch, pulling what it is currently the introduction page into a Write good questions hub page. Should we do the same on develop? If so, I could do this on Wednesday.

AdamChrimes:
@sarawilcox that would be good then we could release this change by Tom independently to the rest of the changes.

AdamChrimes:
@tomdoughty please could you rebase this with `develop`? Will also need a change log entry (I can add this if you want)

tomdoughty:
> @tomdoughty please could you rebase this with `develop`? Will also need a change log entry (I can add this if you want)

The rebase was a bit messy... I think I have got everything in there correctly but maybe worth @davidhunter08 checking I haven't broken anything.

davidhunter08:
> > @tomdoughty please could you rebase this with `develop`? Will also need a change log entry (I can add this if you want)
> 
> @AdamChrimes The rebase was a bit messy... I think I have got everything in there correctly but maybe worth @davidhunter08 checking I haven't broken anything.

 I've removed the class `nhsuk-u-secondary-text-color` from the search results item description.

Other than that, it looks good to me. Nice work.

davidhunter08:
@tomdoughty @AdamChrimes could you have a look at doing an empty search?

When I do an empty search it brings back 279 results, when it should bring back 0 results.

Screenshot 2020-01-07 at 14 38 05

AdamChrimes:
Getting the same as @davidhunter08 with the search results, seems to be working fine on the `develop` branch. @tomdoughty might be caused by some of the JS refactoring somewhere?

tomdoughty:
@AdamChrimes @davidhunter08 Good spot on the empty param handling... added a fix for that 👍

PR #485, opened 2020-01-03
Hub page checks

sarawilcox:
Update hub page content on Design System, Styles, Components and Patterns pages
Amend production and prototyping pages


davidhunter08:
Problems rebasing with `redesign/nav` so added changes manually into `redesign/nav`,

PR #517, opened 2020-01-22
Nunjucks markup cleanup

AdamChrimes:
- Use double quotes (") instead of single quotes (') in the Nunjucks markup. This is common practice and allows us to use single quotes in content without having to escape them.
- Use booleans instead of strings for true and false statements


sarawilcox:
@AdamChrimes We use single quotes as apostrophes but also double quotes for quotations in content. Is that OK?

PR #526, opened 2020-01-23
Fix/broken GitHub links

davidhunter08:


AdamChrimes:
Just need rebasing but looks good

PR #529, opened 2020-01-24
Fix menu button alignment for Safari

davidhunter08:


AdamChrimes:
This should be removed when we release the frontend library https://github.com/nhsuk/nhsuk-frontend/pull/581

PR #574, opened 2020-02-18
Update nhs number pattern with error example

davidhunter08:
Add error summary to NHS number pattern with errors example.


AdamChrimes:
Just needs rebasing with `master` and a CHANGELOG entry adding under 3.0.6

PR #1029, opened 2021-03-09
Service manual team page

davidhunter08:
## What

Add a service manual team page.

## Why

We want to have an official list we can point to for who is on the team.

## Checklist

- [x] Tested against the [NHS.UK frontend testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Code follows the [NHS.UK frontend coding standards](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] Version number has been increased in `package.json` (using [SEMVER](https://semver.org/))
- [x] CHANGELOG entry
- [ ] Page updated date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # NHS digital service manual Changelog
 
+## 3.13.0 - Unreleased
+
+:new: **New features**
+- Add service manual team page
+
 ## 3.12.1 - 08 March 2021
 :wrench: **Fixes**
 
@@ -8,7 +13,7 @@
 ## 3.12.0 - 08 March 2021
 
 :new: **New features**
-- Changed the name of the contribution section to community and contribution
+- Change name of the contribution section to community and contribution
 
 :wrench: **Fixes**
 
--- a/app/styles/app/_footer.scss
+++ b/app/styles/app/_footer.scss
@@ -25,3 +25,24 @@
   }
 
 }
+
+.app-footer__list {
+  margin-bottom: nhsuk-spacing(2);
+
+  @include mq($from: desktop) {
+    margin-bottom: nhsuk-spacing(4);
+  }
+}
+
+.app-footer__text {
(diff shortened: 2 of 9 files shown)


davidhunter08:
image

henocookie:
Maybe link to each individual's GitHub profile? Then people interested could see the things we are all working on easier.

AdamChrimes (approved):
Looks good, pending content sign off

sarawilcox:
I think this is good from the point of view of transparency but wonder what people will want from us here. Are we able to use test this? I'm wondering what people want apart from our names. What was the user need? 

We know that people value being able to contact an individual for support. I suspect that, if we're giving names, people will want to know how to contact us, so how do we want them to do that? By email or public Slack? Do we want people to contact us individually or is it better if they ask their questions in the public Slack so other people can respond? (Mindful of not making a rod for our own backs.)

How does this relate to the Get in touch page?

I'm not sure about linking from wording "Built by" and then just giving the current team. It's been built by past team members too. I wonder if we're better to have the team info on the Get in touch page, where we say to get in touch with the team. That may be better than creating another page. 

Content-wise, could you use en-dashes please, Dave? (pending accessibility testing).

davidhunter08:
I’m not sure about adding links/details for individuals on the page, as we want to steer people towards working in the open.

But there is a need for users ‘to be able to contribute in private as well as public’.

My assumption for this is that users can/will look to find out who is on the team, then message that person on the service manual slack. This is would be good to test @amyj2110 😊.

davidhunter08:
An option could be to remove the "Built by" and instead add a link to the footer links (see screenshot). I've also added a link to the 'Support' section on the home page.

Screenshot 2021-03-09 at 18 30 15

@sarawilcox @AdamChrimes

PR #1042, opened 2021-03-12
add alt text and a caption to the review date image

AdamChrimes:
## Description

add alt text and a caption to the review date image


sarawilcox (approved):
Thanks Adam.

PR #1045, opened 2021-03-15
Update reassure-users-that-a-page-is-up-to-date.njk

sarawilcox:
Moved text about content updates between review dates up the page after feedback during content 2i.


DomBaker (approved):
Reviewed and approved.

PR #1049, opened 2021-03-16
install latest version of the frontend library

AdamChrimes:
## Description

install latest version of the frontend library; remove IE specific classes and SVG fall back logo, import the Search autocomplete directly


tomdoughty (approved):
LGTM

PR #1194, opened 2021-05-14
install latest frontend library and update dependencies

chrimesdev:
## Description

install latest frontend library and update dependencies


davidhunter08 (approved):
LGTM

PR #1221, opened 2021-05-25
Review

sarawilcox:
## Description
Restructure page and amend nav on other pages


chrimesdev:
New pages will need adding to the sitemap so that it can be searched (and indexed)

sarawilcox:
> New pages will need adding to the site map so that it can be searched (and indexed)

Doing that now. Only 1 new page, I think.

karlgoldstraw (approved):
Checked all links, images and text on page and check nav links. 
All seems fine from my point of view.

PR #1245, opened 2021-06-15
Update placeholder colour to darker grey to increase colour contrast

mcheung-nhs:
## Description
Update the search input field placeholder colour to the darker grey to pass colour contrast levels, as per the frontend library:
https://github.com/nhsuk/nhsuk-frontend/blob/7dc973d2f5d2ae788edb9625bde72b879b4969bc/packages/components/header/_header.scss#L187

### Related issue
none

## Checklist

- [x] Tested against the [NHS.UK frontend testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Code follows the [NHS.UK frontend coding standards](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry
- [ ] Page updated date

### DIFF
--- a/app/styles/app/_autocomplete.scss
+++ b/app/styles/app/_autocomplete.scss
@@ -80,7 +80,7 @@
     }
 
     &::placeholder {
-      color: $color_nhsuk-grey-2;
+      color: $color_nhsuk-grey-1;
       font-size: $nhsuk-base-font-size;
     }
   }


chrimesdev (approved):

chrimesdev:
Thanks @mcheung-nhs

cjforms:
I believe that this relates to the 'search' placeholder in the top bar of the site.

Although it's probably better to have a legible search placeholder than an illegible one, it's still a placeholder being uses as a label which is a horrible idea. Somehow we can sort of get away with it in the search box, because people sort of know that a box in the top right corner is likely to be a search box. It's still a really bad idea for all the reasons why placeholders as labels are a bad idea, and the correct fix is to *put the label outside the box* (in this case, put it on the blue area to the left of the box.

PR #1408, opened 2021-08-24
use the correct Service Header example on the Header components page

chrimesdev:
## Description

use the correct Service Header example on the Header components page as it was using the default header twice


DomBaker (approved):
Reviewed and approved

PR #1428, opened 2021-09-03
use anchor link instead of button for start now

chrimesdev:
## Description

Start now link should be an anchor tag styled as button so that it is semantically correct.

Also removed an empty folder.


DomBaker (approved):
Reviewed and approved

PR #1429, opened 2021-09-03
update dependencies to the latest versions

chrimesdev:
## Description

Update dependencies to the latest versions, including some breaking changes.

### To do 

- [x] Double check the dependency breaking changes do not break any of the SM functionality


DomBaker (approved):
Reviewed and approved

PR #1624, opened 2021-11-15
Fix care card example links

henocookie:
Removed links to live NHS.UK pages from the new care card pattern examples (for now)


DomBaker (approved):
reviewed and approved.

PR #1836, opened 2022-12-05
Added a line about how to run tests locally to README

ColinBeeby-Developer:
## Description
Added a line to the README file giving instructions on how to run tests locally

### Related issue
None

## Checklist

- [ ] Tested against the [NHS.UK frontend testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [ ] Code follows the [NHS.UK frontend coding standards](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry
- [ ] Page updated date

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Clone the repo: `git clone https://github.com/nhsuk/nhsuk-service-manual.git nhs
 
 Run the project in development mode `npm run watch` and visit <a href="http://localhost:3000">http://localhost:3000</a>.
 
+Run automated tests locally with `npm run test`.
+
 ## Running the application via the browser with Gitpod
 
 Before running Gitpod, you must <a href="https://github.com/apps/gitpod-io/installations/new">install the Gitpod.io application on your GitHub account</a>.


sarawilcox (commented):
Looks good to me from a content point of view.

andymantell (approved):
I think perhaps we might want to look at the contents of this readme generally and see whether it's fit for purpose. This addition is fair in the sense that someone coming fresh to this might not know to run this command. But if they don't know about this, then there's a whole bunch of other process they are probably also missing. E.g. the main thing that's going to happen when they run that command is that the code will be linted - do they know what this means, etc etc. If they're running tests it's presumably because they've changed something, so do they also know about git and feature branching...

This readme probably could / should be the technical side of the contribution guidelines here: https://service-manual.nhs.uk/community-and-contribution/develop-component-pattern. Or perhaps link out to a page on how to make contributions.

Happy to accept this anyway as it's "correct". Just a slight nag in my head that we need to be clear who the readme is for I suppose, and if we want to document the process we should actually do that properly end to end. Particularly given the situation with an all new team coming on board after a bit of a hiatus.

Issue #1830, opened 2022-11-25
Add back to top code to A to Z page pattern


sarawilcox:
We mention the back to top link in the A to Z pattern but don't include the code in the example.

PR #1902, opened 2023-08-22
Add back to top code to A to Z page pattern

roshaanbajwa:
## Description

Add back to top code to A to Z page pattern

### Related issue

https://github.com/nhsuk/nhsuk-service-manual/issues/1830


the-nathan-smith (approved):
Works fine for me in Gitpod

sarawilcox (approved):
Looks good to me. We should perhaps change the date at the bottom of the page to "August 2023" to show that something significant on the page has changed. Can you do that?

PR #1956, opened 2024-04-24
Change 'open in new window' to 'a new tab'

frankieroberto:
This more accurately describes what the link does, now that most browsers have tabs.

Also follows the convention set by the GOV.UK Design System.


anandamaryon1 (approved):
Good shout, looks good to me.

PR #1964, opened 2024-05-07
Update radio with hint example to call fieldset correctly from radios

### LINKED ISSUE #1962: Radio with hint example uses incorrect markup
The [radios with hint example](https://service-manual.nhs.uk/design-system/components/radios#radios-with-hints) seems to be using incorrect or out of date markup.

It’s using `call` with fieldset, rather than passing a fieldset object to radios like all the other examples do. The resulting markup doesn’t have the hint associated with the fieldset like it should.

edwardhorsford:
## Description
The existing [radio with hint example](https://service-manual.nhs.uk/design-system/components/radios#radios-with-hints) uses `call` with Fieldset, which I think is an older pattern for using this component. By using it this way, the hint doesn't get semantically linked with the fieldset as it should. It also uses two separate divs for each line of the hint, which isn't the markup I would have expected.

This updates the example in all the places it exists to follow all the other radio examples.

I think there's a wider question of whether this hint is a bit long for an example, but for now I think it's better to update the macro so the correct markup for hints is shown.

### Related issue
Fixes #1962 

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/views/content/how-to-write-good-questions-for-forms/use-filter-questions-to-route-users.njk
+++ b/app/views/content/how-to-write-good-questions-for-forms/use-filter-questions-to-route-users.njk
@@ -68,23 +68,25 @@
   <div class="app-example">
     <strong class="nhsuk-tag app-example__heading">Example</strong>
 
-    {% call fieldset({
-      legend: {
-        "text": "Do you know your NHS number?",
-        "classes": "nhsuk-fieldset__legend--m"
-      }
-      }) %}
-      {{ hint({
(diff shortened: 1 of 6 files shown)


frankieroberto (approved):
Checked this by running it locally. All looks good to me. 👍 
 
Screenshot 2024-05-14 at 17 10 02

Thanks @edwardhorsford!

roshaanbajwa (approved):
looks good to me

PR #1966, opened 2024-05-10
Clarify breadcrumb params

frankieroberto:
this updates the Nunjucks documentation for Breadcrumbs to clarify that the top-level `text` and `href` params should be used for the parent page, and not the current page. 

This is because, as the guidance says, "You don't need to show the current page in the breadcrumb because this information is in the H1."

If you do not include the parent page as top-level params, but instead only within the `items` array, the result is identical on desktop views, but on mobile views will result in a back arrow but no text, and an empty link with the hidden text "Back to".

### DIFF
--- a/app/views/design-system/components/breadcrumbs/macro-options.json
+++ b/app/views/design-system/components/breadcrumbs/macro-options.json
@@ -4,7 +4,7 @@
 			"name": "items",
 			"type": "array",
 			"required": true,
-			"description": "Array of breadcrumbs item objects.",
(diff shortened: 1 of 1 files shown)


frankieroberto:
I discovered this by making the mistake myself, and then wondering why our breadcrumb was broken on mobile. Fixed now! 😊 

|Before| After|
|---|---|
|breadcrumb-before | breadcrumb-after |

edwardhorsford (approved):

roshaanbajwa (approved):

PR #1967, opened 2024-05-14
Removes duplicate back links in side nav on design system

anandamaryon1:
## Description

A mobile backlink was showing on the sidebar, even on desktop, it duplicates functionality from the breadcrumbs, so it's been removed.

| Before | After |
|--------|--------|
| image | image |

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/views/whats-new/updates.njk
+++ b/app/views/whats-new/updates.njk
@@ -10,6 +10,25 @@
 
 {% block bodyContent %}
 
+<h2>TBC 2024</h2>
+<table class="nhsuk-table">
+  <caption class="nhsuk-table__caption nhsuk-u-visually-hidden">Updates to the service manual in MONTH 2024</caption>
+  <thead class="nhsuk-table__head">
+    <tr class="nhsuk-table__row">
+      <th class="nhsuk-table__header" scope="col" style="width: 25%;">Section</th>
+      <th class="nhsuk-table__header" scope="col">Update</th>
+    </tr>
+  </thead>
+  <tbody class="nhsuk-table__body">
+    <tr>
+      <td class="nhsuk-table__cell">Design System</td>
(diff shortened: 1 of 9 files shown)


app/views/whats-new/updates.njk
```diff
@@ -10,6 +10,25 @@
...
+    <tr class="nhsuk-table__row">
+      <th class="nhsuk-table__header" scope="col" style="width: 25%;">Section</th>
+      <th class="nhsuk-table__header" scope="col">Update</th>
+    </tr>
+  </thead>
+  <tbody class="nhsuk-table__body">
+    <tr>
+      <td class="nhsuk-table__cell">Design System</td>
+      <td class="nhsuk-table__cell">
+        <p>Removed duplicate links in design system side navigation</p>
```
frankieroberto, line 26: Up to you, but I'm not sure this is enough of a meaningful update to warrant a mention here? Feels more like a minor UX fix to the service manual interface? Slight risk that it might be confusing to mention it as people might imagine that it's an update to a side navigation component (which we don't have as part of nhsuk-frontend)?

frankieroberto (approved):
Nice! Must admit I hadn't spotted this, but now that you've pointed it out, the duplication of the breadcrumb and back link is 🤦.

app/views/whats-new/updates.njk
anandamaryon1: Agree, it could confuse and it's not an important enough update for users.

anandamaryon1: fixed [5926ed9](https://github.com/nhsuk/nhsuk-service-manual/pull/1967/commits/5926ed9da66d9eef21402f12f9fffd0ae1472d00)

frankieroberto (dismissed):

PR #1979, opened 2024-05-28
Add guidance for headings with captions

edwardhorsford:
## Description
Adds guidance for using headings with captions. The content is copied from GOV.UK but I've noted captions can be used below the heading (example) as well as above.

There might want to be some future work looking at the margins where it's used below the heading - but I think a good start is to document what we have and have parity with GOV.UK.

Screenshot 2024-05-28 at 11 07 40

### Related issue
[Existing backlog issue](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/178)

## Checklist
- [x] CHANGELOG entry

### DIFF
--- a/app/views/design-system/styles/typography/index.njk
+++ b/app/views/design-system/styles/typography/index.njk
@@ -90,7 +90,7 @@
     htmlOnly: true
   }) }}
 
-  <p>If the caption should be considered part of the page heading, you can also nest the caption within the <code>{{ "<h1>" | escape }}</code>.</p>
+  <p>If the caption should be considered part of the page heading, you can also nest the caption within the <code>{{ "<h1>" | escape }}</code>, using the appropriate heading and caption classes.</p>
 
   {{ designExample({
     group: "styles",
--- a/app/views/design-system/styles/typography/index.njk
+++ b/app/views/design-system/styles/typography/index.njk
@@ -79,6 +79,26 @@
     htmlOnly: true
   }) }}
 
+  <h2 id="headings-with-captions">Headings with captions</h2>
+
+  <p>Sometimes you may need to make it clear that a page is part of a larger section or group. To do this, you can use a heading with a caption above or below it.</p>
+
+  {{ designExample({
+    group: "styles",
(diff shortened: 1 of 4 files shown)


frankieroberto (approved):
Happy with this, as it documents an otherwise-undocumented feature.

As a follow-up PR, I wonder if it’s worth updating the examples to be more realistic ones, which might better illustrate when the caption should be considered part of the h1?

app/views/design-system/styles/typography/index.njk
```diff
@@ -79,6 +79,26 @@
     htmlOnly: true
   }) }}
 
+  <h2 id="headings-with-captions">Headings with captions</h2>
```
sarawilcox, line 82:
```suggestion
  <h3 id="headings-with-captions">Headings with captions</h3>
```

app/views/design-system/styles/typography/index.njk
```diff
@@ -79,6 +79,26 @@
...
+  <p>Sometimes you may need to make it clear that a page is part of a larger section or group. To do this, you can use a heading with a caption above or below it.</p>
+
+  {{ designExample({
+    group: "styles",
+    item: "typography",
+    type: "headings-with-caption",
+    htmlOnly: true
+  }) }}
+
+  <p>If the caption should be considered part of the page heading, you can also nest the caption within the <code>{{ "<h1>" | escape }}</code>.</p>
```
sarawilcox, line 93:
```suggestion
  <p>If the caption should be considered part of the page heading, you can also nest the caption within the <code>{{ "<h1>" | escape }}</code>, using the appropriate heading and caption classes.</p>
```

app/views/design-system/styles/typography/index.njk
edwardhorsford: good catch, thanks! I copied from GOV.UK design system, but they have headings as their own page.

frankieroberto (approved):

sarawilcox (approved):
All good.

PR #1990, opened 2024-06-18
Scope width overrides

edwardhorsford:
## Description
The service manual overrides the default page width - in some cases styling existing design system classes, in some cases scoping them under app specific classes.

This causes pages to look out of alignment in the page templates section:

Screenshot 2024-06-18 at 16 04 53

This adds extra scoping so the revised width is only used where an app specific class is used. This is added to the service manual header and footer, but doesn't get used by the page template examples so those now look correct.

Arguably a better solution might be for nhs-frontend to support setting the page width like GOV.UK does - that could come as a followup pr.

### DIFF
--- a/app/styles/app/_container.scss
+++ b/app/styles/app/_container.scss
@@ -20,15 +20,13 @@ $service-manual-container-size: 1100px;
 
 // Apply container size to header and footer components
 
-.nhsuk-header__navigation-list {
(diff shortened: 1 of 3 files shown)


anandamaryon1 (approved):
This looks like a useful fix for now. @edwardhorsford would you be able to add an issue for your proposed fix on the nhsuk-frontend please?

PR #1992, opened 2024-06-19
Update to latest Frontend version

roshaanbajwa:
## Description
Update to latest Frontend version


martinsmyllie (approved):
The version updates are minor point version changes. Have we checked that these are non-breaking?

roshaanbajwa:
> The version updates are minor point version changes. Have we checked that these are non-breaking?

Yep the changes have been tested, we've increased the minor semver version to reflect the changes in the new nhsuk frontend version

PR #2034, opened 2024-09-16
Update back link position guidance (from bottom to top)

frankieroberto:
This updates the back link guidance to recommend that generally the back link should appear at the top-left of the page, which is what the majority of services seem to do.

It also explicitly says that it should appear before the `<main>` tag in the HTML so that the skip link will skip past it.

Previously the guidance here suggested that some services included the back link at the bottom of the content. However both of the services mentioned ([111 online](https://111.nhs.uk) and [[Choose if data from your health records is shared for research and planning](https://your-data-matters.service.nhs.uk/)](https://your-data-matters.service.nhs.uk)) seem to have since moved their back link to the top-left.

The current rationale for having the back link at the bottom is:

> That's because we don't want to suggest to people who use a screen reader that they leave the page prematurely.

However, possibly making sure that the back link is skippable using the skip link is enough to mitigate this?

The current guidance is also contradicted by the [page template guidance](https://service-manual.nhs.uk/design-system/styles/page-template) which suggests putting back links in the `outerContent` at the top of the page.

The GOV.UK Design System back link has similar guidance:

> Always place back links at the top of a page, before the `<main>` element. Placing them here means that the ‘Skip to main content’ link allows the user to skip all navigation links, including the back link.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/views/design-system/components/back-link/index.njk
+++ b/app/views/design-system/components/back-link/index.njk
@@ -63,7 +63,7 @@ type: "default"
 
 <p>Generally, the back link should go at the top left of the page. Within the HTML it should appear before the <code>&lt;main&gt;</code> tag. This is so that <a href="/design-system/components/skip-link">skip link</a> skips past the back link to the main content.</p>
 
-<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader.</p>
+<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader: is the page read out in a logical order?</p>
 
 <div class="nhsuk-inset-text app-wcag-inset-text app-wcag-22" id="wcag-interact-back-links" role="note">
   <span class="nhsuk-u-visually-hidden">Information: </span>


frankieroberto:
@anandamaryon1 @sarawilcox this one isn’t dependent on the upcoming NHS.UK Frontend release, however I think it’s worth it going out before or at at same time as that, given it will also mention the back link?

app/views/design-system/components/back-link/index.njk
```diff
@@ -61,8 +61,9 @@ type: "default"
   <p>Do not pre-populate if the information is no longer valid, or when pre-populating would be a major safety or security concern. This is to comply with <a href="https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html">WCAG 2.2 success criterion 3.3.7 Redundant Entry (W3C)</a>.</p>
 </div>
 
-<p>Consider where you put the back link. Do not put it close to other links or buttons where it might distract users
-  from what they need to do. Also think about people who use a screen reader. </p>
+<p>Generally, the back link should go at the top left of the page. Within the HTML it should appear before the <code>&lt;main&gt;</code> tag. This is so that <a href="/design-system/components/skip-link">skip link</a> skips past the back link to the main content.</p>
+
+<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader. </p>
```
edwardhorsford, line 66: "Also think about people who use a screen reader." - what's this trying to say?

frankieroberto: Not quite sure, it's a bit vague! I guess it's implying thinking about the order in which things are read out?

sarawilcox, line 66:
```suggestion
<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader.</p>
```

sarawilcox (approved):
Made a tiny tweak.

app/views/design-system/components/back-link/index.njk
anandamaryon1: > Also think about people who use a screen reader, is page read out in a logical order?

Maybe this?

edwardhorsford: with a colon or hyphen rather than comma? or else rephrase?

edwardhorsford: Ah I didn't see it existed in the previous content too. I think the line is vague and it's unclear what you're meant to do with it.

app/views/design-system/components/back-link/index.njk
```diff
@@ -61,8 +61,9 @@ type: "default"
   <p>Do not pre-populate if the information is no longer valid, or when pre-populating would be a major safety or security concern. This is to comply with <a href="https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html">WCAG 2.2 success criterion 3.3.7 Redundant Entry (W3C)</a>.</p>
 </div>
 
-<p>Consider where you put the back link. Do not put it close to other links or buttons where it might distract users
-  from what they need to do. Also think about people who use a screen reader. </p>
+<p>Generally, the back link should go at the top left of the page. Within the HTML it should appear before the <code>&lt;main&gt;</code> tag. This is so that <a href="/design-system/components/skip-link">skip link</a> skips past the back link to the main content.</p>
+
+<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader.</p>
```
frankieroberto, line 66: Suggestion from @anandamaryon1 and @edwardhorsford:

```suggestion
<p>If you put the back link in a different position, do not put it close to other links or buttons where it might distract users from what they need to do. Also think about people who use a screen reader: is the page read out in a logical order?</p>
```

app/views/design-system/components/back-link/index.njk
frankieroberto: Updated to:

> Also think about people who use a screen reader: is the page read out in a logical order?

frankieroberto: Updated to:

> Also think about people who use a screen reader: is the page read out in a logical order?

frankieroberto:
@sarawilcox @anandamaryon1 @edwardhorsford thanks for the comments (and thanks Sara for spotting the paragraph in the research section! 😊).

I think this is good to be merged now? Although not sure whether you want it merged into `main` or the release branch?

PR #2140, opened 2025-04-03
Update link to Complete multiple tasks pattern

frankieroberto:
We now have a version of this within the NHS design system, so can link there instead of to the GOV.UK one.


sarawilcox (approved):
Oh, well spotted. Thanks for doing that.

PR #2228, opened 2025-07-07
Add guidance on missing data in tables

frankieroberto:
## Description

The GOV.UK Content design guide says that table cells (apart from the top left one) should not be empty for accessibility reasons. Instead, some text should be added.

To make this more distinguishable from regular text, this could use the dark grey colour if necessary (depending on context - might not be needed if other cells contain numbers?). 

We already have an (undocumented) `nhsuk-u-secondary-text-color` modifier, so could use that.

## Screenshot

Screenshot 2025-07-15 at 18 21 39

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 :new: **New features**
 
-- Add new guidance on displaying missing data in tables
+- Add guidance on displaying missing data in tables
 
 ## 7.12.0 - 7 July 2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 - Add new guidance on displaying missing data in tables
 
-
 ## 7.12.0 - 7 July 2025
 
 :new: **New features**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # NHS digital service manual Changelog
 
+## 7.13.0 - TBC
+
+:new: **New features**
+
+- Add new guidance on displaying missing data in tables
+
+
 ## 7.12.0 - 7 July 2025
 
 :new: **New features**
--- a/app/views/design-system/components/table/index.njk
+++ b/app/views/design-system/components/table/index.njk
@@ -75,7 +75,7 @@
     <li>"Not known"</li>
   </ul>
 
-  <p>To distinguish the text from other table content, use the <code>nhsuk-u-secondary-text-color</code> modifier to make it dark grey.</p>
+  <p>To distinguish the text from other table content, use the <code>nhsuk-u-secondary-text-color</code> modifier class to make it dark grey.</p>
 
   {{ designExample({
     group: "components",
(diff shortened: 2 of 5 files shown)


frankieroberto:
@sarawilcox I think those are just examples (taken from the content guide for GOV.UK rather than the GOV.UK Design System). I guess it could be context-dependent a bit?

I’m not sure whether "Not known" or "No data" is better as a default example, any thoughts? I went with "No data" in the the pull request (but forgot to update screenshot) as that’s what the the GOV.UK content guide says, but maybe we could list both?

In our case (Record a vaccination), the vaccine product name could be missing for a number of reasons, such as because it wasn’t entered into the GP system for some reason, or it was but wasn’t made available through our API, or because the patient was vaccinated abroad and doesn’t have a record of which product it was and so the GP had to leave that field blank.  We won’t know which reason it is, so we’d have to use a generic label that covers all those.

Agree that the dark grey "secondary text colour" is pretty close to the regular text colour! It's the same as used in hint text. I guess it’s a trade off between maintaining good contrast against the page background and it being distinguishable from regular text!

anandamaryon1:
This is a really useful addition. Have discussed this on previous projects. 

Suggest we remove 'if this is helpful' for the grey text, since if you can see it, it is helpful to visually distinguish the missing data cells. 

Also could be useful to include example wording for 'not answered' or 'no provided' or similar, for when an optional question has been left empty. -- Or perhaps this is more specific to summary lists and should be included there instead?

sarawilcox:
Screenshot 2025-07-10 at 12 50 24

I've gone for a "Do not" rather than a "should" and have clarified that leaving the top left cell empty is something you can do, not the default. I've also said that the phrases we've included are examples, not a fixed list. People may want to use other wording (e.g. the ones @anandamaryon1 was suggesting for summary lists).

Does this work?

sarawilcox:
I'm assuming we have to include the full stop in Spikevax JN.1.

cjforms:
﻿Thanks Sara.I definitely prefer your new version to mine. Caroline Jarrett @cjformsOn 10 Jul 2025, at 12:54, Sara Wilcox ***@***.***> wrote:﻿sarawilcox left a comment (nhsuk/nhsuk-service-manual#2228)
Screenshot.2025-07-10.at.12.50.24.png (view on web)
I've gone for a "Do not" rather than a "should" and have clarified that leaving the top left cell empty is something you can do, not the default. I've also said that the phrases we've included are examples, not a fixed list. People may want to use other wording (e.g. the ones @anandamaryon1 was suggesting for summary lists).
Does this work?

frankieroberto:
@sarawilcox hmm a lot of the product names contain full stops, like “Comirnaty Omicron XBB.1.5”.

Or could use “Abrysvo” which is an RSV vaccine product.

sarawilcox:
Thanks @frankieroberto . Perhaps use the RSV vaccine because it has a short, simple name (for a vaccine) :-)

anandamaryon1:
Just make some tiny (hopefully non-controversial) content tweaks

anandamaryon1 (approved):

sarawilcox:
Merged into https://github.com/nhsuk/nhsuk-service-manual/pull/2236. We can close this PR.

PR #2229, opened 2025-07-07
Add guidance for secondary text colour

frankieroberto:
We currently have a `nhsuk-u-secondary-text-color` modifier class which isn’t documented in the Service Manual.

This adds some basic guidance. Wasn't sure whether it should be in Typography or Colours, but it felt more natural in the latter.

## Screenshot

Screenshot 2025-07-07 at 16 44 02

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS digital service manual Changelog
 
+## 7.13.0 - TBC
+
+:new: **New features**
+
+- Add guidance on using `nhsuk-u-secondary-text-color` modifier class
+
 ## 7.12.0 - 7 July 2025
 
 :new: **New features**
--- a/app/stylesheets/app/_colour-swatch.scss
+++ b/app/stylesheets/app/_colour-swatch.scss
@@ -21,22 +21,24 @@
   border-radius: 50%;
   display: inline-block;
   height: nhsuk-spacing(6);
-  margin-right: nhsuk-spacing(2);
+  margin-right: 0;
   vertical-align: middle;
   width: nhsuk-spacing(6);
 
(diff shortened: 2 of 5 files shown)


frankieroberto:
@anandamaryon1 I’ve more-or-less got this working without switching the whole table to flexbox. We were already using absolute positioning for the swatch at mobile and tablet sizes, so I switched to using that at all sizes, allowing the padding to then be used to line up the text content:

Screenshot 2025-07-15 at 19 13 18

This would also fix this minor existing glitch I spotted at a specific screen width:

| Before | After |
|--------|-------|
| Screenshot 2025-07-15 at 19 11 29 | Screenshot 2025-07-15 at 19 11 37 |

It could do with a bit more testing though, and maybe the spacing could do with a bit more tweaking.

If we add the modifier class to the table in this way, do you think that would mean we should remove this extra paragraph below:

463308908-517e0bab-d2d8-4ca1-9eb9-83dc434f8359

Or is it worth having both?

anandamaryon1:
Nice one @frankieroberto I'll do some testing this afternoon. 

I think it's good to have both the explainer paragraph and the quick reference in the table.

edwardhorsford:
Nice one @frankieroberto - seems some good improvements.

frankieroberto:
> Nice one @frankieroberto - seems some good improvements.
> 
> I think I might be tempted to mention it from _both_ pages.

@edwardhorsford  Which pages do you mean? This PR adds it to the Colours page, and the other PR [adds it to Tables](https://github.com/nhsuk/nhsuk-service-manual/pull/2228). I think we’d do both, but are there any other pages you were thinking of?

PR #2245, opened 2025-08-15
Update icon guidance

paulrobertlloyd:
## Description

For the next breaking v10 release of NHS.UK frontend, inline SVG icons are provided to components via an `nhsukIcon` macro. This means there is now a single source of truth for our icon set, and no longer a separate store of SVG files which may fall out of sync.

This change however means we need to update our guidance page for icons. Some of the icons are no longer provided by this macro (they are provided by CSS) while links to SVG files for the remaining icons now return a 404 page. Some of the guidance is also out of date.

However, we can use the new macro in the service manual and output the same SVG code for users to copy and paste, as well as show them how they can use the macro to output icons if they are using Nunjucks.

▶️ [View updated icons guidance](https://nhsuk-service-manual-pr-2245.herokuapp.com/design-system/styles/icons)

Screenshot of updated icons guidance page.

### Updates

- Replaces the table with separate code examples for each icon. I have combined the information in this table with later guidance about how these icons should be used
- Adds guidance about the sizing class modifier
- Updates guidance about how to include icons with the Nunjucks macro 
- Fixes some inconsistent and duplicate styles used for code examples

### Right chevron in a circle

This PR currently excludes the chevron in a circle icon. Upon reviewing this guidance, it seemed odd that we have 2 very similar icons, used for very similar purposes. I am proposing consolidating these 2 icons in https://github.com/nhsuk/nhsuk-frontend/pull/1541

### Related issues

- https://github.com/nhsuk/nhsuk-frontend/pull/1028
- https://github.com/nhsuk/nhsuk-frontend/pull/1521

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 - Update footer component and examples
 - Update guidance for using Nunjucks in production code on NHS.UK frontend v10.x
+- Update guidance for icons as we now provide inline SVG icons via nhsukIcon macro
 - Update links to blog post about question protocol
 
 ## 7.13.0 - 28 July 2025
--- a/app/javascripts/design-example.mjs
+++ b/app/javascripts/design-example.mjs
@@ -14,7 +14,9 @@ export class DesignExample extends Component {
     this.hiddenClass = 'js-hidden'
 
     this.tabs = this.$root.querySelectorAll(`.${this.tabClass}`)
-    this.examples = this.$root.querySelectorAll('.code-snippet__preformatted')
+    this.examples = this.$root.querySelectorAll(
+      '.app-code-snippet__preformatted'
+    )
     this.closeButtons = this.$root.querySelectorAll('.app-button--close')
     this.copyButtons = this.$root.querySelectorAll('.app-button--copy')
     this.iframe = this.$root.querySelector('iframe')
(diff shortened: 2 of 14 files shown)


anandamaryon1:
Great, thanks @paulrobertlloyd 

Pushed an update to colour guidance for discussion. Does this help at @sarawilcox?

Not sure what happened with the commit, seems to be lost in the merge, but the new paragraph is there…

sarawilcox:
I've made some minor content tweaks. A few questions.

I put in a sentence top of page about following relevant guidance elsewhere in the design system on use of icons. Is that overdoing things?

Do you think people will expect to find something about the expander + on the icons page? Do we need to mention it (even tho it's not an icon any more)?

We now have a section about colour in icons plus the accessibility para which mentions colour contrast too. I'm wondering about removing the accessibility section. I think we may be better mentioning accessibility (e.g. colour contrast) in the most appropriate context, rather than in a separate section.

Do we still need all the links to further reading about SVGs? We probably needed them when we first published the page. Worth keeping still?

sarawilcox:
@paulrobertlloyd , I can make most of the changes above but wonder what you think about at least mentioning the expander + and - on this page, even though they're now taken care of in the CSS, in case people come to the page expecting them here?

colinrotherham (approved):

frankieroberto (approved):

PR #2252, opened 2025-08-26
Add color rename guidance

frankieroberto:
Add guidance to CHANGELOG and Colours page about the renaming from 'color' to 'colour'.


colinrotherham (commented):
Mind updating these pages too please?

* **app/views/whats-new/index.njk**
* **app/views/whats-new/updates.njk**

Hopefully they're still both in sync with the changelog

PR #2255, opened 2025-08-27
Remove form tags from component examples and make sure others have action attributes

edwardhorsford:
This pr removes `form` tags from all component examples. These are not needed for the examples and may cause issue if users use the 'copy code' button as they probably don't want a form tag.

For any remaining `form` tags (page examples), it adds an `action` attribute to match #2250.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 :wrench: **Maintenance**
+
 - Added form tags to pattern page examples
 - Removed form tags from component examples
 


colinrotherham (approved):

colinrotherham:
Sorry about the linting issue, it only shows up with an NHS Choices account over on Azure:

```console
Checking formatting...
[warn] CHANGELOG.md
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

I'm hoping to resolve this for contributors in https://github.com/nhsuk/nhsuk-service-manual/pull/2256

colinrotherham (approved):
Spot on, thanks @edwardhorsford

PR #2266, opened 2025-09-01
Fix path to new nhsuk-frontend.min.js

frankieroberto:
Discovered a typo in the Upgrade guide for the new path of the precompiled JavaScript.

It is correct in the `CHANGELOG`.


colinrotherham:
>There's also an incorrect closing ~`</body>`~ `</head>` if you can sort that too. Sorry!

Wrong repo, I've sorted it 😅

* https://github.com/nhsuk/nhsuk-frontend/pull/1558

PR #2274, opened 2025-09-04
Nhs number pattern updates - WIP for review

sarawilcox:
## Description

Update NHS number pattern. 

## Related issue

[NHS number pattern issue in community backlog](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/109). 

## To do
- [x] Draw up list of short-term changes - done https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/109#issuecomment-3201142771
- [x] Fix issues
- [ ] Run by stakeholders in Find your NHS number service and PDS
- [ ] Ask PDS team to check PDS page too
- [ ] Outstanding - fix issue with <span class="nhsuk-u-nowrap">987 654 3210</span> not working on desktop @anandamaryon1 
- [ ] 2i checks
- [ ] code review

## Checklist

- [x] CHANGELOG entry
- [ ] [Whats new page](https://service-manual.nhs.uk/whats-new) entry - not yet


app/views/design-system/components/hint-text/default/index.njk
```diff
@@ -1,6 +1,6 @@
 {% from 'hint/macro.njk' import hint %}
 
 {{ hint({
-  text: "Your NHS number is a 10 digit number that you can find on any letter the NHS has sent you, for example, 458 777 3456"
+  text: "This is a 10 digit number, like <span class=\"nhsuk-u-nowrap\">987 654 3210</span>, that you can find on an NHS letter, prescription or in the NHS App"
```
colinrotherham, line 4: You'll need this `text` to be `html` for the HTML to work

```suggestion
  html: "This is a 10 digit number, like <span class=\"nhsuk-u-nowrap\">987 654 3210</span>, that you can find on an NHS letter, prescription or in the NHS App"
```

app/views/content/how-to-write-good-questions-for-forms/use-filter-questions-to-route-users.njk
```diff
@@ -69,16 +69,15 @@
     <strong class="nhsuk-tag app-example__heading">Example</strong>
 
     {% set hintHtml -%}
-      <p class="nhsuk-u-margin-bottom-2">This is a 10 digit number, like 485 777 3456.</p>
-      <p>You can find it on any letter the NHS has sent you, on a prescription or by logging in to a GP practice online service.</p>
+      <p class="nhsuk-u-margin-bottom-2">Conditions include asthma, dementia, diabetes, cancer or heart disease</p>
```
colinrotherham, line 72: Mind if we review these?

If definitely necessary, the class should be moved to `hint.class` and the `hint.html` shouldn't have the paragraph inside, moving to `hint.text`

edwardhorsford: I don't think the class is even needed - I think it's left over from this previously having two paragraphs and them being too far apart. Since this is now a single line, we can get rid of the set block entirely and use `hint.text` as Colin suggests.

colinrotherham:
@sarawilcox I've fixed the conflicts and all the updates from production are applied too

sarawilcox:
Hi, I've gone with a 999 number which doesn't meet mod 11 as more people have fed back that 999 nos are more obviously dummy nos. @anandamaryon1 for info.

I'm now going to run the content updates by the Find your NHS number service and the PDS team. 

We still need to add info about the Nunjucks variable as mentioned in Ed's comment.

sarawilcox:
I've sent the prototype to PDS and FYNN teams for comment.

frankieroberto:
@sarawilcox @anandamaryon1 @colinrotherham spotted a minor bug in this PR on the example using a legend as the page heading, where the wrong class was added, so I’ve fixed it in https://github.com/nhsuk/nhsuk-service-manual/pull/2274/commits/90c7f71789178701cac9127542921df15e790439

This is quite a common mistake because it’s a bit confusing, I wonder if it’s worth explicitly calling out on the page?

Something like:

> When making a legend the page heading, add the `nhsuk-fieldset__legend--l` class to the legend.
>
> When make a label the page heading, add the `nhsuk-label--l` class to the label.

The examples do show this (in both the HTML and Nunjucks versions), so perhaps it’s not necessary, but I wonder if that’s easily missed?

sarawilcox:
Happy to take your advice on this, @frankieroberto @colinrotherham @anandamaryon1 .

colinrotherham:
Similar to https://github.com/nhsuk/nhsuk-service-manual/pull/2340#issuecomment-3501166044, can I ask about `inputmode="numeric"` for NHS numbers?

The example number **999 123 4567** uses spaces but only numbers can be entered:

<br>

Numeric keyboard for NHS number

anandamaryon1:
Similar to https://github.com/nhsuk/nhsuk-service-manual/pull/2340#issuecomment-3501166044, can I ask about `inputmode="numeric"` for NHS numbers?

The example number **999 123 4567** uses spaces but only numbers can be entered:

Ah good spot, can't even enter a decimal point on iOS 🤔. 

We'll need to rethink the 'whole number' example then. Probably something much simpler, but not a suffix like height in cm… Age perhaps?

anandamaryon1:
> Just bear in mind that it only suppresses the link styling, it's still a clickable link

`point-events: none` too?

Or is this all a bit over the top and hacky/risky?

edwardhorsford:
If apple respects the meta tag, add it for this page? (I have a feeling I had to add similar 10 years ago for the passport service)

anandamaryon1:
Ok seems like we have two viable options: 

1. add the meta tag to the example page, and add some guidance about adding it to pages that display NHS numbers (but with caution - it'll apply to the whole page and any other numbers / phone numbers that are not marked up with tel:)
2. add an new class to wrap NHS numbers that removes the link styling

I'm not sure which is better, I suppose 1 (meta tag) is the less hacky solution.

PR #2277, opened 2025-09-08
Clean up markup for confirmation page and panel

edwardhorsford:
## Description

Updates the confirmation example to use double space indentation.

Other updates:

- Prefer `set` over inline html
- Swap panel to using `text` param

### DIFF
--- a/app/views/design-system/patterns/confirmation-page/default/index.njk
+++ b/app/views/design-system/patterns/confirmation-page/default/index.njk
@@ -49,7 +49,7 @@ previewLayout: design-example-wrapper-full
       }) }}
 
       <p>
-        <a href="#">Tell us about your experience using this service (opens in a new tab)</a>
+        <a href="#">Tell us about your experience using this service</a>
       </p>
 
     </div>
--- a/app/views/design-system/patterns/confirmation-page/default/index.njk
+++ b/app/views/design-system/patterns/confirmation-page/default/index.njk
@@ -8,41 +8,51 @@ previewLayout: design-example-wrapper-full
 {% from "summary-list/macro.njk" import summaryList %}
 
 {% block content %}
-    <div class="nhsuk-grid-row">
-        <div class="nhsuk-grid-column-two-thirds">
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-two-thirds">
 
-            {{ panel({
-                titleText: "Booking complete",
-                html: "We have sent you a confirmation email"
-            }) }}
+      {{ panel({
+        titleText: "Booking complete",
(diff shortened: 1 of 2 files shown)


app/views/design-system/patterns/confirmation-page/default/index.njk
```diff
@@ -8,41 +8,51 @@ previewLayout: design-example-wrapper-full
...
+
+      <h2 class="nhsuk-heading-m">Your appointment details</h2>
+
+      {% set siteLocationHtml %}
+        St George's Pharmacy<br>
+        46 St George's Rd,<br>
+        Elephant and Castle,<br>
+        London<br>
+        SE1 6ET<br>
+        <a href="#">Map and directions (opens in a new tab)</a>
```
colinrotherham, line 27: Can we make it open in a new tab, or change the text?

```suggestion
        <a href="#" rel="noreferrer noopener" target="_blank">Map and directions (opens in a new tab)</a>
```

edwardhorsford:
I've updated to:
* remove the link entirely. I don't think it's necessary for the example and raises other questions
* remove punctuation from the displayed address

app/views/design-system/patterns/confirmation-page/default/index.njk
```diff
@@ -8,41 +8,50 @@ previewLayout: design-example-wrapper-full
...
+            },
+            value: {
+              text: "Thursday 15 June at 9:10am"
+            }
+          }
+        ]
+      }) }}
+
+      <p>
+        <a href="#">Tell us about your experience using this service (opens in a new tab)</a>
```
colinrotherham, line 52: Are we keeping this one or removing it?

```suggestion
        <a href="#">Tell us about your experience using this service</a>
```

edwardhorsford: @sarawilcox preference?

frankieroberto, line 52: Should we add the link class? It’s not technically needed, but the other link has/had it so we should be consistent?

```suggestion
        <a href="#" class="nhsuk-link">Tell us about your experience using this service (opens in a new tab)</a>
```

colinrotherham: See https://github.com/nhsuk/nhsuk-service-manual/pull/2277#pullrequestreview-3196269394 for a bit more info

sarawilcox: Better to remove "opens in a new tab". That seems to me to follow WCAG guidance more closely.

PR #2289, opened 2025-09-12
Fix to previous javascripts path

frankieroberto:
We made a small mistake in the v10 upgrade guide by saying that the javascripts path in the templates used to be `nhsuk-frontend.min.js` but it actually used to be `nhsuk.min.js`.

We've also realised that the path prefix is not necessarily `/javascripts/` but can vary depending on your setup (eg the NHS prototype kit uses `/nhsuk-frontend/`). So this makes that clearer by using `<JAVASCRIPTS-PATH>` to hopefully make it clear that that bit isn't part of the upgrade guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS digital service manual Changelog
 
+## Unreleased
+
+:wrench: **Maintenance**
+
+- Update javascript path examples in version 10 upgrade guide
+
 ## 8.0.2 - 12 September 2025
 
 :wrench: **Maintenance**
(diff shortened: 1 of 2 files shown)


frankieroberto:
> Do you mind doing the usual `&lt;` and `&gt;` escaping?

@colinrotherham good catch! Done in 46bd3bf46188838ecfe1cbe5aaf7503826d6128c

colinrotherham (approved):

frankieroberto:
@colinrotherham added brief note to changelog too.

PR #2308, opened 2025-09-19
Use camelCase for all input names

frankieroberto:
Generally we use camelCase for input names rather than hyphens, as it helps to distinguish the 2, and when used in the NHS Prototype kit or other javascript based frameworks, the camelCased properties can be accessed using the dot notation instead of square brackets (`data.fullName` instead of `data['full-name']`).

We didn't do this consistently though, so this switches all the remaining input examples to use camelCase.


frankieroberto:
@sarawilcox this could be something to tackle as part of #2287?  It's not a huge deal, but we noticed it as a minor issue in the training course when someone used the address example.

PR #2352, opened 2025-10-27
Use medium legend for smaller checkboxes and radios

frankieroberto:
Previously the example for smaller radios used a small legend, but the example for smaller checkboxes used a large legend.

This changes them both to medium size.


anandamaryon1:
Thanks, well spotted @frankieroberto. 

Yes, maybe `--m` is best for these examples, albeit that many services will change them as needed based on context.

PR #2354, opened 2025-10-27
Fix HTML issues in update guide

### LINKED ISSUE #2353: "Updating to version 10" guide issue
## What is the issue?

I followed the guide (here: https://service-manual.nhs.uk/design-system/guides/updating-to-v10#page-template) but the main menu wasn't acting as it was supposed to. It wasn't activating the JavaScript to collapse into the "more" menu on smaller screens. It would just scroll sideways.

I noticed that the script the guide tells you to put into the footer:
`<script type="module"> import { initAll } from '/<JAVASCRIPTS-PATH>/nhsuk-frontend.min.js' initAll() </script> 
`
It was throwing the error: 
`Uncaught SyntaxError: Unexpected identifier 'initAll'`

## The fix
To get it to work, I amended the script to read as follows:
```
<script type="module"> import { initAll } from '/includes/js/nhsuk-frontend-10.1.0.min.js';
    initAll();</script>
```
(Added semi-colons and put the initAll() on a new line)

Just wanted to flag this as it was frustrating so it may be worth updating the guide

frankieroberto:
Fixes an issue with incorrect nesting of `<pre>` and `<code>`, which may affect some browsers.

Resolves #2353

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -97,3 +97,31 @@ jobs:
         with:
           name: Javascript unit tests coverage
           path: coverage
+
+  sonar:
+    name: Sonar analysis
+    runs-on: ubuntu-latest
+    needs: [build, tests]
+
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+      SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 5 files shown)


colinrotherham (approved):
Thanks for this! Approving if you're confident we've caught them all

One other thing, think we should adopt the `.app-code` styles?

E.g. Using `tabindex="0"` and `:focus` styles to allow arrow key scrolling

```html
<pre class="app-code"><code tabindex="0" class="app-code__container">
```

Image

colinrotherham:
Only mentioned it because I added `tabindex="0" to ours and scrolling suddenly worked perfectly

We just lacked the focus styles

But yeah can be another issue

PR #2367, opened 2025-11-03
Update the Text input examples and guidance

### LINKED ISSUE #2283: Add examples of Select, Text input and Textarea with the label set at the page heading
@edwardhorsford and I noticed when teaching a prototyping course today that it’s tricky to explain to people how to add page with a single [Textarea](https://service-manual.nhs.uk/design-system/components/textarea) question whether the label is set as the page heading, as there isn’t an example for one that can be easily copied.

When opening the Nunjucks macro options for the Textarea, the `isPageHeading` isn't documented, as that is part of the separate nested Label component, and nested object params aren't currently shown (see #2282).

Even if we did list the `isPageHeading` option, the class to add (eg `nhsuk-label--l`) likely wouldn't be listed, and the Nunjucks params options are quite buried anyway.

Instead I think it makes sense to show examples for Select, Text input and Textarea where the label is set to be the page heading (all the others already do this).

This could either replace the existing example, or we could show both options, with some guidance on when to use either one.

See for example the [GOV.UK Texarkana guidance](https://design-system.service.gov.uk/components/textarea/) which has the first examples setting the label as the page heading, but a later section "If you’re asking more than one question on the page" which shows it not set as the page heading.

frankieroberto:
This updates the default Text input example to set the label as the page heading and to use the `large` font size. This makes it consistent with the other component examples.

Later down the page I’ve added some guidance on when to set the label as the page heading, with another example showing the label that is not the page heading.

Both examples also use a question phrasing for consistency with other examples.

Content inspired partly by the equivalent [Text input guidance on GOV.UK design system](https://design-system.service.gov.uk/components/text-input/).

Partly fixes #2283

## Screenshots

| Before | After |
|-----|----|
| Screenshot 2025-11-03 at 22 39 21 | Screenshot 2025-11-03 at 22 39 12 |

### Guidance

Screenshot 2025-11-03 at 22 42 29

Screenshot 2025-11-03 at 22 42 02

### DIFF
--- a/app/views/design-system/components/text-input/index.njk
+++ b/app/views/design-system/components/text-input/index.njk
@@ -31,7 +31,7 @@
   <p>Give it a visible label. Do not use placeholder text for a label as it vanishes when users click on the text input.</p>
   <p>Align labels above the text inputs they refer to. Labels should be short, direct and written in sentence case. Do not use colons at the end of labels.</p>
 
-  <h3>If you’re asking one question on the page</h3>
+  <h3>If you're asking one question on the page</h3>
 
   <p class="rich-text">If you are asking just <a href="/content/how-to-write-good-questions-for-forms/get-the-questions-into-order#begin-prototyping-with-1-thing-per-page">1 question per page</a> as we recommend, you can set the contents of the <code>&lt;label&gt;</code> as the page heading. This is good practice as it means that screen reader users will only hear the contents once.</p>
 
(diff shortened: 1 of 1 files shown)


sarawilcox:
Yes, I think we've got the coded example covered. I'll tweak the content as you suggest.

sarawilcox:
In fact, I'm including the links out to:
- GOV guidance on labels and legends from section on 1 question per page
- our question page pattern from section on more than 1 question per page

sarawilcox:
Hi @frankieroberto , I've made the content updates but haven't added the coded examples for:
- 1 qu per page
- multiple questions

I wonder if you'd mind doing that, please, if you think it will help? (I struggle a bit with that.)

They're in this branch: https://github.com/nhsuk/nhsuk-service-manual/tree/nhs-number-pattern-updates.

frankieroberto:
@sarawilcox ok, pushed to the other branch in this commit: https://github.com/nhsuk/nhsuk-service-manual/pull/2274/commits/76229d500e82f04dd3557d32ad4df4517e6c3f29

sarawilcox:
Brill, thank you!

PR #2382, opened 2025-11-20
Add guidance for sortable table

sarawilcox:
## Description

First stab at guidance for the sortable table. For the sortable table working group. 

Feel free to comment or add to this. 

## Checklist

- [x] Version number has been increased in `package.json` (using [SEMVER](https://semver.org/))
- [x] CHANGELOG entry


cjforms:
First draft so some quick comments:

1. You've got a stray 'should' in this: "<p>[What types of data should not be sorted?]</p>" And to be honest, I'm not entirely clear what the guidance says in this case.
2. Another question: why is is #basic-tables plural when the other types of table are singular in their tags?
3. And another question: what do you suggest people do with data for the general public instead of using a table?
4. I don't really understand "Do not let users sort a table when the data must be viewed as is". What sort of data must be viewed 'as it is'? Is there any example?

colinrotherham:
I've switched to a new preview branch:

```console
npm install --save github:nhsuk/nhsuk-frontend#50ec9a5c8
```

All working again

sarawilcox:
Hi @frankieroberto and @edwardhorsford , would you have a mo to look at the updated guidance please? You can see it here: https://nhsuk-service-manual-pr-2382.herokuapp.com/design-system/components/table#sortable-table.

Is what I've said about data attributes correct? Note: I haven't worked out how to get the code about data-sort-values to display in the browser, for example, where it says 30 November 2026. 

Frankie, you said in Slack that the default is ascending but the example seems to me to be descending. Is that right? 

Are we going to allow people to change the default? 

Is what I've written about prefixes "a" and "the" OK? I think people may ask how they do that? Is there anything we can say?

frankieroberto:
> Frankie, you said in Slack that the default is ascending but the example seems to me to be descending. Is that right?
> 
> Are we going to allow people to change the default?

It’s defaulting to ascending for me (when you click a new column heading)?  I can’t remember why we picked that as the default though! (I guess you’re more often interested in seeing the biggest numbers?) It’s not yet configurable, but I’ll work on that.

frankieroberto:
@sarawilcox @edwardhorsford I’ve had another go at updating the guidance for changing the custom sort order. It now includes another example showing 3 columns with different types of `data-sort-value` values. I think this is probably needed, as otherwise it’s not super clear how to add these to the HTML or Nunjucks?

Feel free to edit - I might have made it too long?!

sarawilcox:
I like that, @frankieroberto . I think it's much clearer. Thanks for doing it. :blue_heart:

The only thing I'm not clear about (it may be that I'm misunderstanding something) is the bit about default order. In our UR we saw other people get confused about this too so it would be good to make it as clear as poss.

When you come to the page, the childhood vaccination table looks like this, which to me suggests that the default is descending - the MMR column numbers go from large to small and the header shows a down arrow. I assumed that the default example (usually the first on the page) would use default order.

Screenshot 2026-04-08 at 10 38 22

I'm also now wondering why the nations are in the order they're in? Would they be better either as normally quoted: English, Scotland, Wales, Northern Ireland or in alphabetical order? 

With the **Organisations using this service** table, the first column (the nations) shows an up arrow, which says to me ascending, with alphabetical order A to Z. Have I got that right?

Screenshot 2026-04-08 at 10 34 44

sarawilcox:
I think this is almost ready to go, but we're waiting to see if the pilot testing brings up any other issues.

frankieroberto:
@sarawilcox ah sorry, there are 2 different meanings of 'initial sort' being described here:

1) the sort order for the table when the page is first loaded.
2) what happens when you click the header for a currently-unsorted column.

For that second scenario, the default is 'ascending', so it'll go from unsorted to ascending, and then if you click it again it'll keep toggling between the 2.

However on @edwardhorsford's suggestion I’ve added an option to change this default on a per-column basis by setting `data-initial-sort="descending"`. This might make sense, for example, if you think that most users will be more interested in seeing the biggest numbers first for that column. They can still switch to smallest-first by clicking again, so all it’s doing is saving a click.


On the first question - what the default sort order should be when the page first loads, that’s up to the user and there’s no default set (they can choose it by setting `aria-sort="ascending"` or `aria-sort="descending"` on one of the columns. It probably depends on the context?


Tricky thing to explain all this though! I’ve done a quick update to try and explain the new `data-initial-sort` option (which isn't working in the preview yet) - does it make sense at all?

frankieroberto:
> I'm also now wondering why the nations are in the order they're in? Would they be better either as normally quoted: English, Scotland, Wales, Northern Ireland or in alphabetical order?

At the moment in the example the initial sort order is descending by MMR vaccination rates, so they’re in that order. I’d made the first column un-sortable to try and include an example of where not all columns are sortable - but maybe you might well want to sort by region name, and so we should find another example? 🤔 

> With the Organisations using this service table, the first column (the nations) shows an up arrow, which says to me ascending, with alphabetical order A to Z. Have I got that right?

Yep, they’re in ascending order by name. "The Christie" comes first as it's sorting on "Christie" with the "the" dropped (as an example of that feature). Does that help, or is it too confusing?

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+
+  <h2 id="sortable-table">Sortable table</h2>
+  <p>A sortable table lets users sort columns in ascending or descending order.</p>
+  <p>Ascending order (up arrow) arranges:</p>
+  <ul>
+    <li>numerical data from the lowest to the highest value</li>
+    <li>text in alphabetical order A to Z</li>
+  </ul>
+  <p>Descending order (down arrow) reverses this.</p>
+  <p>The default is ascending first, but you can change the default.</p>
```
edwardhorsford:
```suggestion
  <p>The default is ascending first, but you can change the default per column.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
   }) }}
 
   <h3 id="when-not-to-use-a-responsive-table">When not to use a responsive table</h3>
-  <p>Do not use a responsive table:</p>
+  <p>Do not use a responsive table if your content is easier to read and understand on a small screen in a non-responsive layout. Use a <a href="#basic-tables">basic table</a> instead.</p>
+
+  <h2 id="sortable-table">Sortable table</h2>
+  <p>A sortable table lets users sort columns in ascending or descending order.</p>
```
edwardhorsford: I wonder if it's worth mentioning this is a progressive enhancement / requires js. / teams still need to provide the initial data with a suitable sort.

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+  <p>Ascending order (up arrow) arranges:</p>
+  <ul>
+    <li>numerical data from the lowest to the highest value</li>
+    <li>text in alphabetical order A to Z</li>
```
edwardhorsford, line 142: @frankieroberto how does it handle numbers as strings?

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+  <p>The default is ascending first, but you can change the default.</p>
+
+  {{ designExample({
+    group: "components",
+    item: "table",
+    type: "sortable"
+  }) }}
+
+  <h3 id="when-to-use-a-sortable-table">When to use a sortable table</h3>
+  <p>Use this table when users need to find and prioritise data, for example in staff systems with large data sets.</p>
```
edwardhorsford: This is only client side, right? so likely without pagination, and likely not large data sets? Not sure 'large data set' needs to be mentioned.

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+  <h3 id="how-to-use-a-sortable-table">How to use a sortable table</h3>
+
+  <p>This table only sorts by 1 column at a time. Users can click on the column header to trigger sorting.</p>
+  <p>Carry out user research to understand what your users need to do with the data and if a sortable table is the best way to meet their needs. Find out:</p>
+  <ul>
+    <li>which columns they need to sort on</li>
+    <li>how much data to display at once</li>
+    <li>how it should be ordered by default</li>
+    <li>whether to place unsortable data such as empty or null values at the start or end of the table</li>
+  </ul>
```
edwardhorsford: This feels a touch detailed for me. I could imagine teams should consider if it's suitable to sort, but I doubt a team should would go in to this much detail for each table of data.

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+
+  <p>Do not allow sorting on columns that cannot be meaningfully ordered, for example free text.</p>
```
edwardhorsford: Unsure that free text is implicitly unsortable - you can still sort alphabetically and that may be correct.

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+  <h4 id="sort-order">Sort order</h4>
+  <p>By default a column will be sorted:</p>
+  <ul>
+    <li>numerically – if the cells contain numbers</li>
+    <li>alphabetically</li>
+  </ul>
+
+  <p>Do not sort data alphabetically if it has an established logical sequence, for example months of the year, January to December.</p>
+
+  <p>You may want to remove common prefixes like "a" and "the", treating them as "stop words", so that "The Whittington Hospital" comes under W, not under T.</p>
```
edwardhorsford: I wonder if this should mention that you can do this without affecting what's visually shown to the user. I know this is somewhat covered in the next sentence, but I'm not sure that it's obvious that's the case - they seem like two distinct things as written.

Perhaps this sentence could run on _after_ the one about sort order.

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+
+  <p>For example:</p>
+  <ul>
+    <li>dates: 30 November 2026 would be <code>2026-11-30</code></li>
+    <li>formatting numbers: 123,456 would be <code>123456</code></li>
+    <li>currencies: £3.50 would be <code>3.5</code></li>
+    <li>prefixes: The Whittington Hospital would be <code>Whittington Hospital</code></li>
+  </ul>
+
+  <p>By default when you click on the column heading for an unsorted column, it will sort that column in ascending order. You can change this by adding <code>data-initial-sort="descending"</code> to the column heading if descending is the more useful initial sort.</p>
```
edwardhorsford:
```suggestion
  <p>By default when you click on the column heading for an unsorted column, it will sort that column in ascending order. You can change this by adding <code>data-initial-sort="descending"</code> to the column heading if descending is the more useful initial sort direction.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +71,96 @@
...
+    <li>medical conditions</li>
+  </ul>
+  <p>If your team is researching filtering, please feed back your findings via our <a href="https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/217">community backlog issue on filtering a list (in GitHub)</a>.
+
+  <h4 id="session-storage">Session storage</h4>
+  <p>Do not add code to automatically save the sorting order. Users may not expect this and it could add clinical risk.</p>
+  <p>A page refresh or navigation must return the page to its default sort order, unless you add functionality to allow users to save their sorting order explicitly.</p>
+
+  <h4 id="server-side-sorting">Server-side sorting</h4>
+  <p>This component provides client-side sorting, which only sorts the data currently visible in the table on the page.</p>
```
edwardhorsford: Should this describe what client side means or that it relies on JavaScript?

edwardhorsford:
Something I'm not sure is covered is how null or empty values get sorted by default, or indeed if we have any opinions on how they should be handled. It's briefly mentioned in UR - but how will they be treated currently?

Eg should they go to the end? should you try to sort them based on something else?

On a previous project where I designed sortable tables, we had config to decide whether unsortables (null values) went to start or end, or were always at the end, regardless of the sort direction. 

We also supported a fallback sort column - so in the case of a tie (two values the same) or empty cell, we'd fall back to a different column and sort based on that. This meant that we could have empty values all sorted to the end, but then arranged alphabetically.

With the `data-sort-direction` you can _somewhat do this, though it might be cumbersome. I don't think it would be possible to set that empty cells always sort to the end and then alphabetically.

app/views/design-system/components/table/index.njk
frankieroberto: @edwardhorsford If the string contains only a number (ie javascript can interpret it using `Number()` - which means `"10.2"` and `"-5"` both work) then it's handled as a number.

If there are any other characters present, then it's treated as text - eg `"£100"` or `"10ml"` or `"USD 10"`. For those scenarios, you should provide the number itself in the `data-sort-value` attribute.

Hopefully the guidance and the example makes this clear, but if not we should make it clearer!

app/views/design-system/components/table/index.njk
frankieroberto: The component does actually support server-side sorting now as an alternative, although only via full page reloads using link tags rather than ajax.

You use the same syntax, but add an `href`:

```
head: {
 items: [
   {
     text: "Name",
     href: "?sort=name",
     attributes: {
       "aria-sort": "ascending"
     } 
   },
   {
     text: "Date",
     href: "?sort=date",
     attributes: {
       "aria-sort": "none"
     } 
   }
 ]
}
```

However the guidance doesn’t currently mention this.

The client-side sorting is still a progressive enhancement of a non-sortable table. The server-side sorting is for where you have pagination.

app/views/design-system/components/table/index.njk
frankieroberto: "Free text" is perhaps jargon too?

Maybe a better examples would be "a random id number" (like NHS number?) or "a comment from a user" or similar?

frankieroberto:
@edwardhorsford on how empty values should be sorted, currently the javascript sorts blanks first if ascending and numerical, and last if ascending and alphabetic.

This is based on this JavaScript:

```js
return typeof valueA === 'number' && typeof valueB === 'number'
        ? valueA - valueB
        : valueA.toString().localeCompare(valueB.toString())
```        

We’ve not given any thought yet to whether this is right or not, or whether it should be configurable.

The current table guidance also tells people not to leave empty cells at all:

Screenshot 2026-04-13 at 16 03 11

If you use "No data" or "Not applicable" then you’d have to use it alongside a `data-sort-value` attribute, and you could choose to either use `"0"`, or `"-1"` if you wanted it to appear first, or `"999999999999"` or something if you wanted it to appear last (although that's a bit of a hack!).

app/views/design-system/components/table/index.njk
edwardhorsford: Sorting by NHS number also seems like something a team may well want to do.

edwardhorsford:
>If you use "No data" or "Not applicable" then you’d have to use it alongside a data-sort-value attribute, and you could choose to either use "0", or "-1" if you wanted it to appear first, or "999999999999" or something if you wanted it to appear last (although that's a bit of a hack!).

I think each of those options only works for _one_ of the sort directions? So `9999999` might work to sort to end - but as soon as you reverse the sort direction they'd be at the start. Probably not a thing for now, but I don't think the current approach would allow someone to decide unsortables were *always* last.

app/views/design-system/components/table/index.njk
frankieroberto: @edwardhorsford oh really? I guess it might be useful in contexts where you're comparing 2 lists...

Hard to think of a good example of where it's unlikely it’d make sense to sort it!

frankieroberto:
> I think each of those options only works for _one_ of the sort directions? So `9999999` might work to sort to end - but as soon as you reverse the sort direction they'd be at the start. Probably not a thing for now, but I don't think the current approach would allow someone to decide unsortables were _always_ last.

Ah good point. Yeah that’s currently impossible. If this comes up as a need we could figure out a way to support it in future. Maybe you'd set `data-sort-always-last="true"` or something, instead of a `data-sort-value`? 🤷

anandamaryon1:
Have just pushed some small updates to the examples:
* Make "Childhood vaccination coverage" example table not sorted by default
* Add row headers to both examples (for accessibility – still need to decide whether all table examples need this, but that's not sortable table related)
* Make "Organisations using the service" sorted initially by amount spent instead of name

Tried **unsuccessfully** to update the frontend version to make the `data-initial-sort` work… (I ran `npm publish-preview` on the frontend branch, copied over the npm install code with SHA and ran it, but the feature still doesn't work.)

anandamaryon1:
@colinrotherham I tried to update the frontend version for this PR, to enable the latest features (eg. `data-initial-sort`) and it didn't work. Any chance you could have a go? 

And maybe advise me how to update a branch preview once it's already been created (I think that's the problem I have, it works first time but not then to pull in new updates, even when regenerating the preview branch and running the new install SHA). See this commit: https://github.com/nhsuk/nhsuk-service-manual/pull/2382/changes/8888fbf942cb97b958dc96a6a36bb61339008bcc.

sarawilcox:
@anandamaryon1 . I tweaked wording re returning users to first page when sorting across paginated pages. 

Screenshot 2026-05-19 at 10 42 09

frankieroberto:
@sarawilcox @anandamaryon1 I’ve added some more draft guidance for the server-side version of the table sorting. It includes an example (with pagination) as that seemed like the easiest way to explain it?!

From a guidance point of view, I think the simplest rule of thumb is:

* if table is paginated, then always use the server-side version
* if table is not paginated, then use JavaScript version

From a Nunjucks/HTML point of view, the main difference is that the server-side version uses `href`/links in the header columns, whereas with the JavaScript version those are added as buttons automatically by the JavaScript (if available).

sarawilcox:
Hi @frankieroberto , I like the example and additional guidance but am not sure about page structure. Let me have a think.

colinrotherham:
NHS.UK frontend preview is now up to date

Fixed the indentation and ~single~ double quotes on the new examples too

Do we need to swap the `data-*` attributes with updated options?

sarawilcox:
Hi @frankieroberto , I think what confused me was that we lost the heading "Sort order". Given that most of the content under the heading "JavaScript version" is about sorting, I feel we need something to highlight that. Just calling it "JavaScript version" threw me off the scent. 

Does this order work better? 

H3 How to use a sortable table
    H4 JavaScript version
          H5 Sort order (including the content on not saving sort order from further down the page) - we don't like using H5s but I feel the sorting section will otherwise get lost under a heading "JavaScript" for people who don't know to expect it there
    H4 Server-side version
    H4 Consider search and filter (I assume that applies to both JS and server-side?)

What do you think?

frankieroberto:
@sarawilcox @anandamaryon1 good suggestion! I've done that in 3d5f431edfcbe155d3e60ce78dcc2f93bc32ec2b - whaddya think?

5 is a lot of heading levels though! 🤔 

This little intro could now contain jump links to the 2 headings? Although we do also already have jump links at the very top of the page...

Screenshot 2026-05-19 at 14 26 58

frankieroberto:
@anandamaryon1 @sarawilcox @colinrotherham another option... we follow the pattern from the Styles section, and separate these out as sub-pages? 😱 

sub-pages

sarawilcox:
I think the structure's OK, even though it's 5 levels deep. 

You could add jump links where you introduce JavaScript and server-side versions. We may not need them though as we go straight into talking about the JavaScript version after that short intro. I'm tempted not to. 

Another option would be to mention at the top of the page (with jump links) that we have 2 versions of the sortable table: JavaScript and server-side. Would people want to know that right at the start?

colinrotherham:
I'm still torn on the names “JavaScript version” and “server-side version” I suppose

In the sense that we don’t describe any of our other components as being the "server-side version"—it's the default?

Perhaps we should consider the `href` as an alternative where sortable tables are not available:

>Do not use sortable tables when paginated because only the current page's table can be sorted. Alternatively, when using Nunjucks you can add the `href` option to each head cell to link to a page sorted on the server.

Wording wise, we chose to say "on the server" for password input

sarawilcox:
I'm not sure about applying the Styles section pattern, @frankieroberto . We have several other components that have multiple sections, e.g. buttons and header. At some point I think we should have a wider review of design system/service manual nav (perhaps when we publish a new side nav component?) but at the mo people seem to navigate the long pages OK with the jump links at the top. (We did test this page.) Mind you, we don't have jump links to the sections on word breaks or displaying missing data.

anandamaryon1:
## Design system review panel review summary and actions

### Component
* Not sure the brackets in the hidden text affect how it’s read out, could save some bytes by removing them? **Action**: investigate and remove if no effect
* **Action**: investigate scrolling container to allow for scrollable sortable tables

### Guidance
* **Action**: Mention in guidance that sortable columns show a double arrow until sorted, and a single arrow after sorting - DONE
* If there is no data in a sorted column’s cell, where will it be sorted? **Action**: check and add to guidance - DONE
* **Action**: mention that this is most suited to data-heavy tables (IGNORE)
* **Action**: mention that it doesn’t work with the responsive table variant - DONE
* Sub-headings of “JavaScript version” and “Server-side version”. **Action**: make these less technical - headings re Sort order and something to do with Multiple pages - DONE

### Examples
* Example is sorted by an unsortable column, this is confusing because a user cannot return to the initial sort order once they’ve sorted. **Action**: Don’t show examples where the default sort order is invisible. Update examples, consider mentioning in guidance

anandamaryon1:
> * Not sure the brackets in the hidden text affect how it’s read out, could save some bytes by removing them? **Action**: investigate and remove if no effect

Tested on VoiceOver and NVDA: the brackets add a slight pause which helps to break up the sentence. Will keep.

anandamaryon1:
> * **Action**: investigate scrolling container to allow for scrollable sortable tables

Work in progress: https://github.com/nhsuk/nhsuk-frontend/pull/1969

sarawilcox:
Hi @anandamaryon1 , I'm making some updates to the sortable table page. I notice that we talk about right aligning numeric data and have a class for that. We also have a section on text alignment in Typography that mentions right aligning some languages (like Arabic) with a right align class. I wondered about linking the 2 sections but they use different classes. What do you think?

sarawilcox:
@anandamaryon1 I'm rewording the content that focused on browser vs server-side and now talks about sorting data on the page and sorting pages of data. Would you mind taking another look at it, please?

sarawilcox:
The section on sorting data on a page feels rather disorganised to me. Is there a way we could help people get their heads round it better by adding some sub-headings? 

Or is there a better order for some of it?

anandamaryon1:
> Hi @anandamaryon1 , I'm making some updates to the sortable table page. I notice that we talk about right aligning numeric data and have a class for that. We also have a section on text alignment in Typography that mentions right aligning some languages (like Arabic) with a right align class. I wondered about linking the 2 sections but they use different classes. What do you think?

I don't think linking is necessary. Although they are both right aligning text, they're for different purposes and different scope.

anandamaryon1:
@sarawilcox I've had a go at adding subheadings (h5's 😭) to the sorting data on this page section. I've tried to make a distinction between overall column header level factors and then the sorting within those columns. Maybe that makes sense? Maybe we could think of better headings? Let me know what you think.

cjforms:
Hi @anandamaryon1 
Haven't been able to look at this for ages for <reasons>. I agree with you that subheads help, but I even more agree with you that when we get to including some H5s then that's a red flag for 'this page is too complicated'. 

1. On sorting - My suggestion is to

- number the two methods
- use verb phrases (1. sort the data on the page / 2. sort pages of data) rather than gerunds (sorting)
- even, consider using the entire phrases 'Method 1: sort the data on the page" / "Method 2: sort pages of data" because I got really muddled with what was a description of what's going on and what was the rest of the typical design system page. 
- swap the two methods because "sort pages of data" is much, much shorter. It's harder for the developers too, which is another story. But I'd got quite lost in 'data on the page' by the time I came across it. 

The description of sort data on the page is a bit confused. For example, the topic of whether the data is sorted alphabetically or numerically is split, with these two extra sentences appearing part-way through it: 
"Do not add code to automatically save the sort order. Users may not expect this and it could add clinical risk.

A page refresh or navigation must return the page to its default sort order, unless you add functionality to allow users to save their sort order explicitly."

2. I had to let the guidance marinate in my head for a bit before I grasped that 'sort data on the page' will only sort the visible data, so will not pull any data from following pages whereas 'sort pages of data' will pull data that's not yet visible to the user. I think it would be better to point this out.  I'd even go so far as to recommend that 'sort data on the page' is better restricted to scenarios where there is only one page of data. 

3.  On missing data: 
There's a stray "should' in this bit: 

"Cells with [missing data](https://nhsuk-service-manual-pr-2382.herokuapp.com/design-system/components/table#missing-data-table) should show "No data" or similar. "

Try this instead: 
If a cell has [missing data](https://nhsuk-service-manual-pr-2382.herokuapp.com/design-system/components/table#missing-data-table) then show "No data" or similar. "
( the 'or similar' does the heavy lifting here)

anandamaryon1:
- [ ] Add example and guidance for scrollable table container: https://github.com/nhsuk/nhsuk-frontend/pull/1969

sarawilcox:
Hi @cjforms , thanks for your comments. Always helpful.

I don't like the H5s either but, given that we've got several kinds of table on this page, it's become long and complex. However, in testing, we found that people expected the sortable table on the tables page and they seem to navigate these long pages OK.

We only use numbered lists when something is done in sequence which doesn't apply here. But you're right that's it's better to use "Sort" than "Sorting" in the heading. I've updated that.

I've left the 2 methods in the order they are as, at this point, we're expecting more people to do the simpler thing of sorting on the page. But @anandamaryon1 , please say if you disagree.

I've reordered things slightly but it's still a bit of a mish-mash. I'm not sure how to improve it further at this stage. 😬

You're right that sorting data on the page should only be where you have 1 page of data. I've tried to make that clear now. I've also dealt with the stray "should".

@anandamaryon1 , would you mind double-checking it, please?

colinrotherham:
@sarawilcox I've updated the NHS.UK frontend preview to the latest version

I've updated the table page to include a reverse example and I've wired up the compact one

app/views/design-system/components/table/index.njk
```diff
@@ -50,11 +83,110 @@
...
+    item: "table",
+    type: "sortable-custom",
+    showJavaScriptToggle: true
+  }) }}
+
+  <h4 id="sort-more-than-1-page-of-data">Sort more than 1 page of data</h4>
+
+  <p>If your table is paginated, do not use the JavaScript-based sorting as this will only sort the current page being shown.</p>
+
+  <p>Instead, use server-side sorting, with the column headings containing links that reload the page with the table sorted by that heading. You will need to:</p>
```
frankieroberto:
```suggestion
  <p>Instead, use server-side sorting, with the column header containing links that reload the page with the table sorted by that column. You will need to:</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -52,7 +52,8 @@
   }) }}
 
   <h3 id="3-or-more-column-table">3 or more column table</h3>
-  <p>If you have numeric data, right align the column header and cells to make it easier to compare numbers.</p>
+  <p>If you have numeric data, right align the column header and cells to make it easier to compare numbers. If using Nunjucks, set the <code>format: numeric</code> macro option. In HTML add the <code>nhsuk-table__cell--numeric</code> class to the column's header and cells.</p>
```
colinrotherham: @anandamaryon1 To complicate things slightly it's:

* `nhsuk-table__cell--numeric` for `<td>` elements
* `nhsuk-table__header--numeric` for `<th>` elements

But take care, `<th>` header cells aren't limited to column headers only

We also use `nhsuk-table__header` in rows with `firstCellIsHeader: true`

app/views/design-system/components/table/index.njk
```diff
@@ -52,7 +52,8 @@
...
+  <p>For non-numeric data that you want to right align, use the <code>nhsuk-u-text-align-right</code> class instead.</p>
```
colinrotherham: Can we better explain how to add classes to cells?

I've seen three different people try to do it this way in recent weeks:

```json5
{
  html: '<span class="nhsuk-u-secondary-text-colour">No data</span>'
}
```

Instead of the correct way:

```json5
{
  text: "No data",
  classes: "nhsuk-u-secondary-text-colour"
}
```

Same advice applies to `nhsuk-u-text-align-right` and other classes

app/views/design-system/components/table/index.njk
anandamaryon1: Whoops, I'll fix now

app/views/design-system/components/table/index.njk
anandamaryon1: How's this? 
Adding: `If using Nunjucks, set it with <code>classes</code>.`

app/views/design-system/components/table/sortable-custom/index.njk
```diff
@@ -0,0 +1,107 @@
+{% from "tables/macro.njk" import table %}
+
+{{ table({
+  caption: "Organisations using the service",
+  captionSize: "m",
+  firstCellIsHeader: true,
+  head: [
+    {
+      text: "Organisation",
+      classes: "nhsuk-u-width-one-third",
```
colinrotherham: In the release version we'll have support for the `width` option

```suggestion
      width: "one-third",
```

app/views/design-system/components/table/sortable-custom/index.njk
```diff
@@ -0,0 +1,107 @@
...
+      text: "Amount spent",
+      format: "numeric",
+      sort: "descending"
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "Warrington and Halton Hospitals",
+        sortValue: "Warrington and Halton Hospitals"
```
colinrotherham: Should we drop the `sortValue` when it's identical? 🤷‍♂️

```suggestion
        text: "Warrington and Halton Hospitals"
```

app/views/design-system/components/table/sortable-custom/index.njk
```diff
@@ -0,0 +1,107 @@
...
+      },
+      {
+        text: "RY2N0"
+      },
+      {
+        text: "18 July 2024",
+        sortValue: "2024-07-18",
+        format: "numeric"
```
colinrotherham: Are we using `format: "numeric"` in place of `align: "right"` here?

In the sense that `sortValue: "2024-07-18"` is correctly ordered already (numeric not necessary)

```suggestion
        sortValue: "2024-07-18",
        align: "right"
```

app/views/design-system/components/table/sortable-custom/index.njk
```diff
@@ -0,0 +1,107 @@
...
+      {
+        text: "£16,821",
+        sortValue: "16821",
+        format: "numeric"
+      }
+    ],
+    [
+      {
+        text: "The Christie",
+        sortValue: "Christie"
```
colinrotherham, line 49: They always refer to themselves as "The Christie", so I felt a bit odd about this one

app/views/design-system/components/table/sortable-custom/index.njk
```diff
@@ -0,0 +1,107 @@
...
+      },
+      {
+        text: "21 August 2022",
+        sortValue: "2022-08-21",
+        format: "numeric"
+      },
+      {
+        text: "No data",
+        sortValue: "0",
+        format: "numeric",
```
colinrotherham: Due to how [`Intl.Collator` sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) works in browsers we'll need the currency here:

```suggestion
        sortValue: "£0",
        format: "numeric",
```

<br>

Or alternatively you can set `format: "string"` where the `sortValue` is unnecessary

```suggestion
        format: "string",
```

app/views/design-system/components/table/sortable-server-side/index.njk
```diff
@@ -0,0 +1,188 @@
...
+  head: [
+    {
+      text: "Name",
+      href: "#",
+      sort: "descending"
+    },
+    {
+      text: "Last log in",
+      href: "#",
+      classes: "nhsuk-u-text-align-right",
```
colinrotherham:
```suggestion
      align: "right",
```

app/views/design-system/components/table/sortable-server-side/index.njk
```diff
@@ -0,0 +1,188 @@
...
+      sort: true
+    },
+    {
+      html: '<span class="nhsuk-u-visually-hidden">Action</span>'
```
colinrotherham: Can we make sure classes go in `classes` to remove the nested `<span>`?

Alignment only needs to be set in the header cell, It's auto-applied to the rest of the column

```suggestion
      text: "Action",
      align: "right",
      classes: "nhsuk-u-visually-hidden"
```

app/views/design-system/components/table/sortable-server-side/index.njk
```diff
@@ -0,0 +1,188 @@
...
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "Abel Pope"
+      },
+      {
+        text: "15 January 2026",
+        classes: "nhsuk-u-text-align-right"
```
colinrotherham: Will only need `align: "right"` once in the header cell

```suggestion
        text: "15 January 2026"
```

app/views/design-system/components/table/sortable-server-side/index.njk
```diff
@@ -0,0 +1,188 @@
...
+      },
+      {
+        html: '<a href="#" class="nhsuk-link">Change<span class="nhsuk-u-visually-hidden"> details for Abel Pope</span>',
+        classes: "nhsuk-u-text-align-right"
```
colinrotherham: Ahh this is very interesting

Should we allow `item.href` on regular table cells too?

Maybe even `item.visuallyHiddenText`?

```suggestion
        href: "#",
        text: "Change",
        visuallyHiddenText: "details for Abel Pope",
        align: "right"
```

app/views/design-system/components/table/index.njk
```diff
@@ -25,6 +53,8 @@
 
   <h3 id="3-or-more-column-table">3 or more column table</h3>
   <p>If you have numeric data, right align the column header and cells to make it easier to compare numbers.</p>
+  <p>If using Nunjucks, set the <code>format: numeric</code> macro option. In HTML add the <code>nhsuk-table__header--numeric</code> class to the column's header and <code>nhsuk-table__cell--numeric</code> class to the cells.</p>
```
colinrotherham: Can we explain exactly where to set `format: "numeric"`?

* Add Nunjucks `format: "numeric"` to table header cells only
* Add HTML `nhsuk-table__header--numeric` class to table header cells
* Add HTML `nhsuk-table__cell--numeric` class to table body cells

Don't forget the double quotes too:

```suggestion
  <p>If using Nunjucks, set the <code>format: "numeric"</code> macro option. In HTML add the <code>nhsuk-table__header--numeric</code> class to the column's header and <code>nhsuk-table__cell--numeric</code> class to the cells.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -33,14 +63,29 @@
...
+
+  {{ designExample({
+    group: "components",
+    item: "table",
+    type: "striped"
+  }) }}
 
   <h2 id="responsive-table">Responsive table</h2>
-  <p>This table has a responsive layout. On large screens it displays as a 3 column table. However, on small screens it stacks vertically.</p>
+  <p>This table has a responsive layout. On large screens it displays horizontally but on small screens it stacks vertically.</p>
```
colinrotherham, line 105: Can we include instructions on Nunjucks `responsive: true`?

We could explain about the `--responsive` CSS class too but every table element also sets its `role` when responsive (to restore what CSS removes), like `role: "rowheader"` or `role: "cell"`, so it's a bit harder to describe

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
   }) }}
 
   <h3 id="when-not-to-use-a-responsive-table">When not to use a responsive table</h3>
-  <p>Do not use a responsive table:</p>
+  <p>Do not use a responsive table if your content is easier to read and understand on a small screen in a non-stacked layout. Use a <a href="#basic-tables">basic table</a> instead. If you have too many columns to fit, use a scrolling table.</p>
+
+  <h2 id="scrolling-table">Scrolling table</h2>
+  <p>Add the scrolling container when your table has many columns and you cannot split it up or use a responsive table.</p>
```
colinrotherham, line 119: Have we decided what the Nunjucks option should be?

Good to explain that it's the `<div class="nhsuk-table-container"> </div>` wrapper for HTML

app/views/design-system/components/table/index.njk
colinrotherham: Just to clarify this now we're using [`Intl.Collator` sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator)

You can now:

1. Set `format: "numeric"` and let the browser sort units (`£`, `ml`, `%`) automatically
3. Set `sortValue` for different units only, e.g. `sortValue: "£0.50"` for `text: "50p"`
3. Set `sortValue` on every cell in the sortable column

It's quite hard to explain so 3) is probably easier

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+    <li>sort data that fits on 1 page</li>
+    <li>sort more than 1 page of data</li>
+  </ul>
+
+  <h4 id="sort-data-on-1-page">Sort data on 1 page</h4>
+
+  <p>The sortable table component uses JavaScript to sort the data in the user's browser. Use it where you have only 1 page of data. Make sure your data has a suitable sort order if JavaScript does not run.</p>
+
+  <h5 id="default-sort-order">Sorting columns</h5>
+  <p>Decide which column you want to sort by when the page first loads. Make sure users can get back to this. Do not load the page sorted on a column which users cannot sort.</p>
```
colinrotherham: What is a column that users cannot sort? Should explain this

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+  <p>By default when you click on the column header for an unsorted column, it will sort that column in ascending order. You can change this by adding <code>sortNext: "descending"</code> in Nunjucks (or in HTML <code>data-initial-sort="descending"</code>) to the column header if descending is the more useful way to sort it initially.</p>
```
colinrotherham:
```suggestion
  <p>By default when you click on the column header for an unsorted column, it will sort that column in ascending order. You can change this by adding <code>sortNext: "descending"</code> in Nunjucks (or in HTML <code>data-sort-next="descending"</code>) to the column header if descending is the more useful way to sort it when clicked.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+  <p>Typically users expect text to sort A to Z by default. With numbers, they often expect them to sort with the biggest number first.</p>
+  <p>Do not allow sorting on columns that cannot be meaningfully ordered, for example comments from users.</p>
+  <p>Do not add code to automatically save the sort order. Users may not expect this and it could add clinical risk.</p>
+  <p>A page refresh or navigation must return the page to its default sort order, unless you add functionality to allow users to save their sort order explicitly.</p>
+
+  <h5 id="sorting-numbers">Sorting numbers</h5>
+  <p>By default, columns will be sorted as text. To correctly order columns containing numbers, set the <code>format: numeric</code> option in Nunjucks. In HTML add the <code>nhsuk-table__header--numeric</code> class to the column header and <code>nhsuk-table__cell--numeric</code> class to the cells).</p>
```
colinrotherham: We're kinda [duplicating the content above](https://github.com/nhsuk/nhsuk-service-manual/pull/2382/changes#r3664658961)?

```suggestion
  <p>By default, columns will be sorted as text. To correctly order columns containing numbers, set the <code>format: "numeric"</code> option in Nunjucks. In HTML add the <code>nhsuk-table__header--numeric</code> class to the column header and <code>nhsuk-table__cell--numeric</code> class to the cells).</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+
+  <h5 id="overriding-default-sort-order">Overriding the default sort order</h5>
+  <p>You can override the default sort order to make the data easier for users to understand. Do this by adding the Nunjucks option <code>sortValue</code> (or in HTML the <code>data-sort-value</code> attribute) to the individual table cells containing either a number or text on which to sort.</p>
+  <p>For example:</p>
+  <ul>
+    <li>dates: 30 November 2026 would be <code>2026-11-30</code></li>
+    <li>prefixes: The Whittington Hospital would be <code>Whittington Hospital</code></li>
+  </ul>
+
+  <p>Do not sort data alphabetically if it has an established sequence, for example months of the year (January to December).</p>
```
colinrotherham: Is this correct?

We can use `"sortValue: "01 January"` to `"sortValue: "12 December"`

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+
+  <p>If a cell contains <a href="#missing-data-table">missing data</a>, say "No data" or similar. By default these cells will be sorted by first letter. With numbers, they will be treated as zero. You can change this by adding a custom <code>sortValue</code>. If using <code>numeric</code> columns with symbols, include the symbol, for example <code>sortValue: "£0"</code>.</p>
```
colinrotherham: Probably stick with "alphabetically" instead of "by first letter"?

They won't actually be treated as numbers (sorted numerically) unless `format: "numeric"` is added so it's probably simpler to go straight to `sortValue` at this point:

```suggestion
  <p>If a cell contains <a href="#missing-data-table">missing data</a>, say "No data" or similar. By default these cells will be sorted alphabetically. To sort numerically, add the Nunjucks option <code>sortValue</code> (or in HTML the <code>data-sort-value</code> attribute), for example <code>sortValue: "£0"</code>. Include symbols or units to sort consistently with other cells.
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+ {{ designExample({
+    group: "components",
+    item: "table",
+    type: "sortable-custom",
+    showJavaScriptToggle: true
+  }) }}
+
+  <h4 id="sort-more-than-1-page-of-data">Sort more than 1 page of data</h4>
+
+  <p>If your table is paginated, do not use the JavaScript-based sorting as this will only sort the current page being shown.</p>
```
colinrotherham: We don't split out JavaScript based sorting anymore

```suggestion
  <p>If your table is paginated, sorting should be applied server-side across all pages. Do not sort the current page only.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+
+  <p>Instead, use server-side sorting, with the column header containing links that reload the page with the table sorted by that column. You will need to:</p>
```
colinrotherham: We don't usually say that links reload pages. They navigate elsewhere.

```suggestion
  <p>Use server-side sorting by adding a link to the column header to navigate to a page with the table sorting already applied. You will need to:</p>
```

app/views/design-system/components/table/sortable-custom/index.njk
sarawilcox: There are a lot of hospitals that may or may not be searched for under The. Especially when people use their short names

- The Lister
- The Whittington 
- The Royal Free
- The Royal London
- The Royal Marsden
- The Hillingdon Hospitals NHS Foundation Trust
- The Princess Alexandra 
- The Queen Elizabeth Hospital 
- The Royal Orthopaedic 
- The Royal Wolverhampton 
- The Dudley Group 
- The Walton Centre 

to name a few

I think generally speaking it's more helpful to list them under the first letter of the word that follows The.

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+
+  <ul>
+    <li>build your own server-side function</li>
+    <li>return users to the first page of new results after sorting</li>
+    <li>test with users</li>
+  </ul>
+  <p>This example shows a paginated table but it does not work in the browser. The functionality would need to be built server-side.</p>
```
colinrotherham, line 231: Can we please build this?

E.g. We can set up multiple examples and set the `href` to jump between them:

* `sortable-server-side/index.njk`
* `sortable-server-side-ascending/index.njk`
* `sortable-server-side-descending/index.njk`

```suggestion
  <p>This example shows a paginated table with sorting applied server-side.</p>
```

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+  {{ designExample({
+    group: "components",
+    item: "table",
+    type: "sortable-server-side",
+    showJavaScriptToggle: true
+  }) }}
+
+  <h4 id="consider-search-and-filter">Consider search and filter</h3>
+  <p>Sometimes it's better to let users search or filter results instead of sorting. Filtering cuts the data down so users can focus on what they want. Search helps them find specific things they need.</p>
```
colinrotherham: Is "cuts the data down" plain English enough?

* When is "sometimes"—when trying to find something?
* Why "instead of sorting"? Use both. E.g. Filter by location, sort by name

```suggestion
  <p>Sometimes it's better to let users search or filter results instead of sorting. Filters reduce the amount of data so that users can focus on what they want. Search helps them find specific things they need.</p>
```

app/views/index.njk
```diff
@@ -86,25 +86,27 @@
           }) }}
         </li>
       </ul>
+      <hr class="nhsuk-u-margin-top-2 nhsuk-u-margin-bottom-5">
```
colinrotherham: Can we use a default section break here?

That way we get responsive margins

(Perhaps adjust the bottom margin on the `<ul>` if it's too large)

```suggestion
      <hr class="nhsuk-section-break nhsuk-section-break--l">
```

app/views/index.njk
```diff
@@ -86,25 +86,27 @@
...
         <div class="nhsuk-grid-column-two-thirds">
           <h2 class="nhsuk-heading-l">
             Coming soon
           </h2>
-          <p>We're always working with the community to improve the service manual. Find out what the team is working on now on our <a href="/community-and-contribution/whats-coming-soon/">What's coming soon page</a> and how you can contribute.</p>
+          <p>We're always working with the community to improve the service manual. Find out what the team is working on now on our <a href="/community-and-contribution/coming-soon/">Coming soon page</a> and how you can contribute.</p>
         </div>
       </div>
-      <hr class="nhsuk-u-margin-top-2 nhsuk-u-margin-bottom-6">
+      <hr class="nhsuk-u-margin-top-3 nhsuk-u-margin-bottom-5">
```
colinrotherham: We should use our default section break spacing here too?

```suggestion
      <hr class="nhsuk-section-break nhsuk-section-break--l">
```

app/views/design-system/components/table/sortable-custom/index.njk
anandamaryon1: Not sure, the `sortVlaue` is a kind of number?

app/views/design-system/components/table/sortable-server-side/index.njk
anandamaryon1: Interestingly, adding the class instead of `html` makes an odd gap where the `th` border bottom is missing too.

app/views/design-system/components/table/index.njk
anandamaryon1: `scrollable: true`? Or did you have some thoughts on it?

app/views/design-system/components/table/sortable-custom/index.njk
colinrotherham: If you want to, the code doesn't need it though

The ISO 8601 date format (`YYYY-MM-DD`) was ordered that way so A-Z sorting still worked

colinrotherham: If you want to keep it, don't forget the server-side example doesn't use it

(It uses text-align instead of numeric format)

app/views/design-system/components/table/sortable-server-side/index.njk
colinrotherham: Ah because the whole cell is visually hidden

Not great really is it 😬

app/views/design-system/components/table/index.njk
colinrotherham: Now that we ditched `sortable: true` it can probably go back to non-adjective `scroll: true`?

Otherwise `responsiveable: true` 😉

app/views/design-system/components/table/index.njk
sarawilcox: We have seen 2 people now get confused that this example doesn't actually work. I think we need to say (perhaps in the example itself) that it's not a working example but an illustration of a table with sorting applied server-side.

app/views/design-system/components/table/index.njk
anandamaryon1: Ah yes, `scroll: true` is nice, saves the world a few superfluous bytes 👌

app/views/design-system/components/table/index.njk
anandamaryon1: Ooh nice, yes, I hadn't quite realised it is that simple…

app/views/design-system/components/table/index.njk
anandamaryon1: But that wouldn't be sorting alphabetically in a user's view. We mean don't sort a column April, August, December etc.

Maybe we should mention months in the custom order section? (but we already mention dates which is similar?)

app/views/design-system/components/table/index.njk
anandamaryon1: @sarawilcox? 

>When is "sometimes"—when trying to find something?

Yes maybe this can be preceded by: When users need to find a specific thing…?

Think we're trying to say: don't assume sorting will replace the need for filtering and searching.

app/views/design-system/components/table/index.njk
```diff
@@ -50,14 +95,117 @@
...
+    <li>build your own server-side function</li>
+    <li>return users to the first page of new results after sorting</li>
+    <li>test with users</li>
+  </ul>
+  <p>This example shows a paginated table but it does not work in the browser. The functionality would need to be built server-side.</p>
+  {{ designExample({
+    group: "components",
+    item: "table",
+    type: "sortable-server-side",
+    showJavaScriptToggle: true
```
anandamaryon1: I'm thinking we remove this, since it doesn't do anything here. In a sense that's the point, but also, I think it's confusing.

app/views/design-system/components/table/index.njk
anandamaryon1: Added new pages and linked them up, so that the example works now. (Used AI, took about 5 tries but saved me some typing and thinking up lots of names)

app/views/design-system/components/table/index.njk
anandamaryon1: Thoughts @sarawilcox ? 

It's repeated but far up the page. I suppose we can just mention the option? eg.

> By default, columns will be sorted as text. To correctly order columns containing numbers, set the format to numeric. 

Maybe with a link back up?

app/views/design-system/components/table/index.njk
anandamaryon1: Will update once frontend updated and new preview pulled in here

app/views/design-system/components/table/index.njk
anandamaryon1: How's this?

> Enable it in Nunjucks by adding the responsive: true option to the table. In HTML add the nhsuk-table--responsive class to the table and set the correct role attribute for each element of the table.

app/views/design-system/components/table/sortable-server-side/index.njk
anandamaryon1: Will wait for `item.href` from frontend

app/views/design-system/components/table/sortable-server-side/index.njk
anandamaryon1: Waiting on frontend preview update, will then implement

app/views/design-system/components/table/sortable-server-side/index.njk
anandamaryon1: Have to leave with `html:`?

app/views/design-system/components/table/sortable-server-side/index.njk
anandamaryon1: Waiting on frontend preview update, will then implement

app/views/design-system/components/table/sortable-custom/index.njk
anandamaryon1: Added the `£`. Not sure what's best though.

anandamaryon1: Maybe adding the symbols and units is more consistent?

app/views/design-system/components/table/sortable-custom/index.njk
anandamaryon1: Ok will swap it out when `align: "right"` is ready

app/views/design-system/components/table/index.njk
colinrotherham: Yeah it's better. For reference, we do this in the frontend changelog for complex HTML:

>If you are not using Nunjucks macros, use the HTML markup from the [table examples in the NHS digital service manual](https://service-manual.nhs.uk/design-system/components/table).

app/views/design-system/components/table/index.njk
colinrotherham: Ah but no, see the `01` and `12` prefixes I've added to the `sortValue`?

Users don't see the padded number prefix but it makes sure they stay in order:

* 01 January
* 02 February
* 03 March
* 04 April
* 05 May
* 06 June
* 07 July
* 08 August
* 09 September
* 10 October
* 11 November
* 12 December

Same approach as for dates, works in non-numeric A-Z order too

app/views/design-system/components/table/index.njk
anandamaryon1: Decided to keep but reformat a little. Users tend to not read the whole page top to bottom so we feel keeping it is helpful for those looking specifically at sortable tables, and it being a small bit of content, not worth cross linking.

app/views/design-system/components/table/index.njk
anandamaryon1: Removed this line, added a month example to the bullets

app/views/design-system/components/table/index.njk
anandamaryon1: Applied your rewording and removed first line as it was ambiguous.

app/views/design-system/components/table/index.njk
```diff
@@ -24,7 +52,12 @@
   }) }}
 
   <h3 id="3-or-more-column-table">3 or more column table</h3>
-  <p>If you have numeric data, right align the column header and cells to make it easier to compare numbers.</p>
+  <p>If you have numeric data, right align the column header and cells to make it easier to compare numbers:</p>
```
colinrotherham, line 62: Is this line confusing now?

It talks about right aligning the column but we also have our new option now:

* `align: "left"` (default)
* `align: "centre"`
* `align: "right"`

They're shortcuts to add the `nhsuk-u-text-align-*` class

Perhaps the guidance should say that setting the numeric format will automatically right align the column, but alignment can also be set for non-numeric values too? With a link to alignment guidance

anandamaryon1: It does go on to say about non-numeric, but I need to update it to use `align: "right"` instead of `classes`

anandamaryon1: How's this?

image

app/views/design-system/components/table/sortable-server-side/index.njk
colinrotherham: If you're still happy with both `item.href` and `item.visuallyHiddenText`?

We can use the latter. I'll open a PR

anandamaryon1:
@colinrotherham hero has disappeared from home page, not sure why, can't see any errors in terminal or console.

anandamaryon1:
Also, notice that the what's new date caption on homepage has swapped to top rather than bottom placement, is the on purpose? Makes more sense at bottom to me.

colinrotherham:
>@colinrotherham hero has disappeared from home page, not sure why, can't see any errors in terminal or console.

Ah I must have taken the preview branch from _before_ we merged the hero changes in

Let's make another one this afternoon

>Also, notice that the what's new date caption on homepage has swapped to top rather than bottom placement, is the on purpose? Makes more sense at bottom to me.

Sorry no, not on purpose—this is just the default from the `heading()` macro

We can change the options to set the caption placement 👀

colinrotherham:
NHS.UK frontend preview updated

@anandamaryon1 Hero is back and I've moved the August 2026 caption from start to end

anandamaryon1 (approved):

sarawilcox:
This is all looking nice. I made a tiny change to the header for the responsive table. I see that you updated it to match the table on the NHS website page which is great but their table header was confusing me so I've tweaked it. Otherwise it's good that we're following their clinically assured content.

The info about Nunjucks vs HTML is much clearer now. Are we going to be adding the 2 options more widely across the design system? It does add more content to the page. I wonder if we should get Laura to test it in upcoming research?

colinrotherham:
>The info about Nunjucks vs HTML is much clearer now. Are we going to be adding the 2 options more widely across the design system? It does add more content to the page. I wonder if we should get Laura to test it in upcoming research?

I've just pushed up https://github.com/nhsuk/nhsuk-service-manual/pull/2382/commits/8a33e31f6a02e441f70e71f7220fdfd161a08e44 to add `variant: "reverse"` guidance

Whether it's the best approach I'm not sure. Our screening teams are widely the same options (but via Jinja)

@anandamaryon1 If you'd like to check the updated options in https://github.com/nhsuk/nhsuk-service-manual/pull/2382/commits/5c0060ca2e8ee71e4f688cf4218ebe7035305f5d please

anandamaryon1:
> I've just pushed up [8a33e31](https://github.com/nhsuk/nhsuk-service-manual/commit/8a33e31f6a02e441f70e71f7220fdfd161a08e44) to add `variant: "reverse"` guidance
> 
> Whether it's the best approach I'm not sure. Our screening teams are widely the same options (but via Jinja)
> 
> @anandamaryon1 If you'd like to check the updated options in [5c0060c](https://github.com/nhsuk/nhsuk-service-manual/commit/5c0060ca2e8ee71e4f688cf4218ebe7035305f5d) please

Thanks, those look good to me @colinrotherham.

I'm not sure on the best approach either. Might be worth some UR as @sarawilcox mentioned. Pros and cons to all approaches… I'll make a ticket to think about it (EDIT: added: https://github.com/nhsuk/nhsuk-service-manual/issues/2636)

colinrotherham (approved):

PR #2395, opened 2025-12-09
SPIKE: Selectable table

frankieroberto:
This seems to be a common-ish pattern in staff-facing services, where there's a table containing rows that have individual actions, but also a checkbox per row you can use to do something to multiple rows at once (edit / delete / activate / etc).

Screenshot 2025-12-09 at 12 30 00

Having a 'Select all' checkbox can also be handy, both for actually selecting all, or for selecting all and then deselecting some, if you need to select most of them. That feature is being explored separately in https://github.com/nhsuk/nhsuk-frontend/pull/1707.

Open questions:

* Should the checkbox and label be in single `<td>` or in separate ones?
* Should the 'Name' header row be the label for the 'Select all' checkbox, or should that have a visually-hidden label and the 'Name' header be separate?
* Should button at the bottom change state if there is nothing selected, or should it remain the same but show some kind of error if nothing selected?
* Is the "Actions" header helpful, or should this be empty or visually-hidden?
* If the table is the main content of the page, should the table caption be the `<h1>`? Or is it better to have a separate `<h1>` and make the table caption smaller (or even absent or visually-hidden)?
* Should the spacing be tighter on the rows?

### DIFF
--- a/app/views/design-system/components/table/with-checkboxes/index.njk
+++ b/app/views/design-system/components/table/with-checkboxes/index.njk
@@ -1,0 +1,65 @@
+{% from "tables/macro.njk" import table %}
+{% from "checkboxes/macro.njk" import checkboxes %}
+
+{% set organisations = [{name: "NHS England", code: "X26"}, {name: "NHS Business Services Authority", code: "T1450"}, {name: "NHS Blood and Tranplant", code: "T1460"}, {name: "UK Health Security Agency", code: "X36"}] %}
+
+{% set rows = [] %}
+
+{% set headerCheckboxHtml %}
+  {{ checkboxes({
+    idPrefix: "organisations",
+    name: "organisations",
+    classes: "nhsuk-checkboxes--small",
+    formGroup: {
+      classes: "nhsuk-u-margin-bottom-0"
+    },
+    items: [
+      {
+        value: "",
+        text: "Name",
+        id: "organisation-select-all",
+        inclusive: true,
+        label: {
+          classes: "nhsuk-label--s"
+        }
+      }
+    ]
+  }) }}
+{% endset %}
+
+{% for organisation in organisations %}
+  {% set nameCheckboxHtml %}
+    {{ checkboxes({
+      idPrefix: "organisations",
+      name: "organisations",
+      classes: "nhsuk-checkboxes--small",
+      formGroup: {
+        classes: "nhsuk-u-margin-bottom-0"
+      },
+      items: [
+        {
+          value: organisation.code,
+          text: organisation.name,
+          id: "organisation-" + loop.index0
+        }
+      ]
+    }) }}
+  {% endset %}
+
+  {% set rows = (rows.push([{ html: nameCheckboxHtml}, {text: organisation.code}]), rows) %}
+{% endfor %}
+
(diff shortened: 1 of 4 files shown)


mmips:
hi @frankieroberto, in one of the past projects (quite complex staff-facing journey) we had a similar tables. the insights we were getting when testing the journeys make me think that the more clarity you can provide, the better. it may depend what users are used to, which patterns they interact with often etc. (usual stuff). i'd think that:
- unless you make it explicit what the top checkbox means, it's up for interpretation and some people may get it, some may not
- button at the bottom - have you considered adding "select to edit" instruction on the top of the table, keeping the button as is (should it be primary button if it's the only one on the page?), giving someone an error when they try to click on the button without selecting?
- i'd keep the header Actions - or Action if it remains only one
- about the H1 - I think it may be ok to leave 'Organisations' as H1, but feels to me that a table could use some more context, e.g.  hint text with 'Select to edit' or similar. not sure what's on the previous screens, maybe it's more self-explainatory with that context
- imo the spacing seems fine, is it the same as proposed in [the service manual](https://service-manual.nhs.uk/design-system/components/table)?

Also, did you get feedback about both checkboxes and 'Edit' links? I'm wondering if it's not confusing for users as you have ultimately two actions doing very similar (the same?) thing. May it be better to have one or the other for simplicity?

roobottom:
Here's how we're using a similar pattern 

screencapture-mya-prototype-11c9c1433d21-herokuapp-site-1-remove-74e895c4f2-2025-12-09-16_16_07

And the markup

```
<table class="nhsuk-table" id="bookings-table">
  <thead class="nhsuk-table__head">
    <tr>
      <th scope="col" class="nhsuk-table__header app-table-select">
        <div class="nhsuk-checkboxes__item nhsuk-checkboxes--small app-table-select-checkbox app-table-select-checkbox--header">
          <input class="nhsuk-checkboxes__input" id="select-all" type="checkbox" name="select-all" value="true">
          <label class="nhsuk-label nhsuk-checkboxes__label" for="select-all"><span class="nhsuk-u-visually-hidden">Select all </span>Dates</label>
        </div>
      </th>
      <th scope="col" class="nhsuk-table__header nhsuk-table__header--numeric">Total bookings (all services)</th>
    </tr>
  </thead>
  <tbody class="nhsuk-table__body">
    <tr class="nhsuk-table__row">
      <td class="nhsuk-table__cell">
        <div class="nhsuk-checkboxes__item nhsuk-checkboxes--small app-table-select-checkbox">
          <input class="nhsuk-checkboxes__input" id="select-date2025-12-09" type="checkbox" name="select-date" value="2025-12-09">
          <label class="nhsuk-label nhsuk-checkboxes__label" for="select-date2025-12-09">9:30am to 1pm on Tuesday 9 December 2025</label>
        </div>
      </td>
      <td class="nhsuk-table__cell nhsuk-table__cell--numeric">56</td>
    </tr>
  </tbody>
</table>
```

It's pretty much a standard NHS table with the addition of custom `app-table-select-checkbox` which also borrows from the standard nhs-checkbox. 

I haven't done much thinking about the markup beyond how we could test the concept. There might be some accessibility no-nos here.

Issue #2384, opened 2025-12-01
Agree a way of calling out updates to design system pages


sarawilcox:
We now have breaking changes in inset text at the top of many design system pages but some pages have now had further, more recent updates.

We need to agree and test a better way of noting latest changes for different audiences. 

Related issue: https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/31

Could we do something like GOV? (See [comment in above GitHub issue](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/31#issuecomment-3462012383).)

anandamaryon1:
An idea for a minimal version of this, where we at least show the current frontend version, and a link to a page where we can explain how to check your current version. 

Would want to explore placing the version info at the top, since the code snippets can be long. But then there's a clash with the macro options details to consider.

<img width="1656" height="1152" alt="Image" src="https://github.com/user-attachments/assets/4eba20d2-76a4-4729-b0cc-dfed79e9d2c0" />

anandamaryon1:
- I need to know the version of the frontend that is used in the design system examples
- - I need to know the earliest version this component works with
- If I copy a code snippet from the design system examples and my local version doesn't match, the code may be wrong
- If my prototype/service is stuck on a specific frontend version, I need to know what I can use in design system
- I need know what guidance has changed and when
- I need to know what I can use and can't for my current version
- I need to know what has changed and when so that I can be sure I'm following best practice
- I need to know which frontend version corresponds to a prototype version
- I need to know that I have to update my prototype version to use the latest release
- I need to know how to update my frontend / prototype version
- I need to know what the latest version of frontend is, when looking at the design system

sarawilcox:
Can we close this one now, @anandamaryon1 ? Design system pages now show the FE version number. 

<img width="748" height="452" alt="Image" src="https://github.com/user-attachments/assets/71ca1e91-3ea0-441c-b43c-d0c8273a35e3" />

anandamaryon1:
GOV design system have added histories for components: https://design-system.service.gov.uk/components/text-input/history/

<img width="857" height="493" alt="Image" src="https://github.com/user-attachments/assets/e648922d-7f0f-4aa2-8d0d-a1d945866a4e" />

<img width="213" height="221" alt="Image" src="https://github.com/user-attachments/assets/54912a53-632f-4f20-9d4f-214a1fdce826" />

PR #2432, opened 2026-01-28
Showing version number in design system

anandamaryon1:
## Description

A basic approach to showing the current frontend version in the design system. To help avoid confusion when users expect code snippets to work in their prototype but their version doesn't match. 

Gone with a basic approach because we don't think it's a massive issue and probably not worth spending too much time on a solution. (it comes up occasionally ~ every few months, and can be resolved fairly easily when users reach out).

Would hope that this can mean we retire the inset text calling out recent updates. The page about checking version can link to changelogs and the guide and explain how to update. 

image

Todo: pull in frontend version from somewhere (npm package?).

localhost_3000_design-system_guides_check-frontend-version (1)

### Related issue

https://github.com/nhsuk/nhsuk-service-manual/issues/2384

## Checklist

- [ ] CHANGELOG entry


anandamaryon1:
Yeah fair point on the capitals… We do sometimes say NHS.UK frontend. 
image


Or more wordy:
image

sarawilcox:
I think they look better/easier to read but see what other people think.

anandamaryon1:
> @anandamaryon1 would it always show the very latest version for every component, or would we somehow tag each component with the version number where it was last updated? 🤔

That would be nice, but for simplicity I was thinking the former. Unless there's a neat automatic way to do that. But I was assuming it'd be tricky and maybe difficult to maintain.

anandamaryon1:
That would be ideal @colinrotherham!

anandamaryon1:
Urgh sorry… Tried to resolve the conflicts by doing a `git rebase origin/main` and it's made a mess.

sarawilcox:
Have made a few tiny tweaks @anandamaryon1 . Otherwise this is fine.

anandamaryon1:
@colinrotherham am I correct in thinking that the version number is currently showing the same on all components? I.e we don't have the version number for when this component was last updated? Is that possible to implement?

colinrotherham:
>@colinrotherham am I correct in thinking that the version number is currently showing the same on all components? I.e we don't have the version number for when this component was last updated? Is that possible to implement?

Yeah it's just whatever version is in the `nhsuk-frontend` package

I'll still need to have a look at this but it's possible for sure

Would we essentially use **macro-options.json** to do this?

* Loop all non-deprecated options
* Pick the highest `version` number

anandamaryon1:
> Would we essentially use **macro-options.json** to do this?
> 
> * Loop all non-deprecated options
> * Pick the highest `version` number

Yes I think that could work. I suppose there are levels to it and to what we would term 'working'. E.g adding a new `caller` option which doesn't directly impact the usage of the component would update the number, but I think we'll have to let that level of detail slide. You've done a lot of updating and aligning options across components recently so shouldn't be a big issue moving forward. It'll still be helpful information.

colinrotherham:
@anandamaryon1 Updated to display the maximum (non-deprecated) NHS.UK frontend version number

colinrotherham (approved):

PR #2494, opened 2026-03-24
Add autocomplete example for date of birth

frankieroberto:
On the Date input page, both examples currently shown are "What is your date of birth?", however neither of them use the `autocomplete` attributes. 

It is currently only mentioned in the Nunjucks macro options:

Screenshot 2026-03-24 at 09 56 21

This updates the page to add a section explaining the autocomplete attribute, and to give an example of how to use it.

Screenshot 2026-03-24 at 10 49 46

I’ve changed the 2 other examples to "What is their date of birth?" as an example of where not to use the autocomplete attribute. However we could pick something more generic like "When was your appointment?" or similar.

See [GOV.UK Date input page](https://design-system.service.gov.uk/components/date-input/) for their equivalent guidance.

### DIFF
--- a/app/views/design-system/components/date-input/default/index.njk
+++ b/app/views/design-system/components/date-input/default/index.njk
@@ -1,17 +1,17 @@
 {% from "date-input/macro.njk" import dateInput %}
 
 {{ dateInput({
-  id: "date-of-birth",
-  namePrefix: "dateOfBirth",
+  id: "appointment-date",
+  namePrefix: "appointmentDate",
   fieldset: {
     legend: {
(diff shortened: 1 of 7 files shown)


sarawilcox (approved):
Looks good to me.

sarawilcox:
Hi @frankieroberto , the content designer who was 2i-ing this got a bit confused by "your" and "their" DOB. It seemed clearer when I suggested using "When was your appointment?" as you suggested above. I'll amend the questions.

colinrotherham (approved):

PR #2515, opened 2026-04-10
Add javascript toggle to components that use progressive enhancement

frankieroberto:
For some of our components (eg Tabs, Character count, Password, conditionally checkboxes, etc), they behave differently if JavaScript is unavailable, however this isn’t always apparent. Browsers also no longer make it easy to quickly turn javascript off (it's buried in menus or developer tools).

This suggestion is for a JavaScript toggle that we can enable for certain component examples.

It'd let you toggle between the 2 versions, to quickly see them, or you could open the currently shown version in a new tab.

(And if javascript is unavailable when viewing the service manual page itself, I guess we'd not show the toggle at all)

A secondary benefit of this would be to help remind people that progressive enhancement is one of the requirements of government digital services (even if this isn't always followed).

Here’s how it might look:

| On | Off |
| ---- | ----|
| Screenshot 2026-06-24 at 00 39 04 | Screenshot 2026-06-24 at 00 39 27 |
| Screenshot 2026-06-24 at 00 38 30 | Screenshot 2026-06-24 at 00 38 49
| Screenshot 2026-06-24 at 00 40 42 | Screenshot 2026-06-24 at 00 41 19 |
 
Thoughts? @edwardhorsford @colinrotherham @michaelgallagher?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS digital service manual Changelog
 
+## TBC
+
+:wrench: **Maintenance**
+
+- Add curly quote example to punctuation section
+
 ## 8.12.1 - 17 June 2026
 
 :wrench: **Maintenance**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 :new: **New features**
 
 - Add option for "all" to checkboxes component
+- Add new reference page for override classes
 
 :wrench: **Maintenance**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add curly quote example to punctuation section
 - Update character count component to mention new data-count-type
+- Update guidance on getting accessibility audits
 
 ## 8.12.1 - 17 June 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## TBC
 
+:new: **New features**
+
+- Add option for "all" to checkboxes component
+
 :wrench: **Maintenance**
 
 - Add curly quote example to punctuation section
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add curly quote example to punctuation section
 - Update character count component to mention new data-count-type
 - Update guidance on getting accessibility audits
+- Update wording about terms of use on Start page pattern
 
 ## 8.12.1 - 17 June 2026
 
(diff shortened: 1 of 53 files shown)


colinrotherham:
Yeah this would be good to demonstrate

We'll need to watch out that the `nhsuk-frontend-supported` class is still set, so various components will continue to apply styling as if its JavaScript is running

In the context of our `nhsuk-frontend-supported` class it might be helpful to demonstrate the (more likely) scenario that a browser is too old to support modern JavaScript

frankieroberto:
@colinrotherham yeah when testing this locally I did get a bit confused initially when the Tabs component still looked like tabs even though it didn't work - because the `nhsuk-frontend-supported` class was still there. I couldn't see a quick way to turn that off from the example pages - it might need a separate template?

colinrotherham:
Yeah it'll have to be a separate template to avoid the mandatory `<script>` 😣

frankieroberto:
I'll leave this as an idea for the team to consider for now - but happy to help if this gets picked up.

I think the components/examples it’d apply to would be:

- Tabs
- Character count
- conditional checkboxes and radios
- 'None' checkbox 
- Password input
- Header (the 'More' menu relies on javascript)

and then in future:

- Sortable table
- Stepper
- 'Select all' checkbox

frankieroberto:
@colinrotherham @anandamaryon1 got this properly working now, and enabled it on all components that use progressive enhancement to make some meaningfully visible changes.

Lemme know what you think!

Could also be useful to demo how the upcoming 'All' checkbox option works with no-JavaScript (ie it gets hidden)?

colinrotherham:
Absolutely magic 🙌

Loved seeing character count (subtle) versus file upload and tabs

frankieroberto:
@colinrotherham glad you like it! It’ll be good for sortable tables and the “all” checkbox too.

One extra reason I think it’s helpful is that Safari and Chrome both seem to have made it harder to quickly toggle JavaScript off and on than it used to be.

@anandamaryon1 @sarawilcox any thoughts?

anandamaryon1:
This is great @frankieroberto, helps to highlight and communicate no-JS considerations. 

I've had a go at improving the responsive layout and making it announce state. I found the aria-current wasn't working. But maybe I've gone too far and it's a bit messy now, probably there's a simpler approach. I also swapped the link to buttons so that they can be pressed with spacebar.

sarawilcox:
Ooh, interesting. Sounds good to me. ❤️

colinrotherham (approved):

Issue #2484, opened 2026-03-18
First draft of guidance for secondary navigation


sarawilcox:
To support:  Secondary navigation
https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/556
- [ ] Prep initial draft
- [ ] Run by Ed
- [ ] Run by working group
- [ ] Add page to site map

PR #2518, opened 2026-04-14
Add section navigation guidance

sarawilcox:
## Description
Prepare draft guidance for section navigation component

### Related issue

First draft of guidance for secondary navigation #2484

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/views/design-system/components/section-navigation/index.njk
+++ b/app/views/design-system/components/section-navigation/index.njk
@@ -1,6 +1,6 @@
 {% extends "layouts/app-layout.njk" %}
 
-{% set pageTitle = "Tabbed navigation" %}
+{% set pageTitle = "Section navigation" %}
 {% set pageSection = "Design system" %}
 {% set subSection = "Components" %}
 {% set pageDescription = "A horizontal list of secondary navigation links to allow users to navigate between pages in a section of a service." %}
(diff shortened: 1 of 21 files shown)


anandamaryon1:
Updated title to section nav, added a frontend preview and added some working examples

app/views/design-system/components/section-navigation/index.njk
```diff
@@ -0,0 +1,111 @@
...
+    <li>using headings</li>    
+    <li><a href="/design-system/components/details">details</a></li>
+    <li><a href="/design-system/components/expander">expanders</a></li>
+    <li><a href="/design-system/components/tabs">tabs</a></li>
+</ul>
+
+<p>Details, expanders and tabs hide sections of content on a page which a user can choose to reveal.</p>
+
+<h2 id="how-to-use">How to use section navigation</h2>
+<p>Section navigation is a horizontal list of links. Use it in a full-width column.</p>
```
anandamaryon1: This intro relegates the vertical option, I wonder whether people could miss it. 

I think we should mention both at the outset, and then talk about each in turn.

sarawilcox: I've made the section on making page titles and headings unique a sub-section of horizontal nav but does it apply to both? Should it come under vertical nav?

anandamaryon1: Yes it's relevant to both

anandamaryon1: But how I see it now, as the first sub-head under how to use, it seems fine

colinrotherham:
@anandamaryon1 I've switched to `direction: "vertical"` from https://github.com/nhsuk/nhsuk-frontend/pull/1886

For small screens with short-enough section navigations, we had a slight flicker when JavaScript kicked in

But to avoid this, we can now add `direction: "horizontal"` to set the initial render direction

This is just a preference (bit like an initial sort order) and JavaScript will change to vertical if it doesn't fit

anandamaryon1:
Pushed some updates:
* add an example showing that the horizontal version collapses on small screens
* added page additions for related search terms

app/views/design-system/components/section-navigation/vertical/index.njk
```diff
@@ -0,0 +1,46 @@
...
+          {
+            href: "#",
+            text: "Vaccinations"
+          },
+          {
+            href: "#",
+            text: "Referrals",
+            badge: {
+              html: "New<span class=\"nhsuk-u-visually-hidden\"> feature</span>"
+            }
```
colinrotherham, line 36: We see this pattern a lot where `visuallyHiddenText` can add a `"Count"` prefix, but for other components that same option could be a suffix instead—it varies

But here we're only needing a suffix (not a prefix) so we have to manually append HTML

Would be good to discuss how we might solve this via a new Nunjucks API:

* Icons and heading captions use `placement` as `"start"`, `"end"` etc
* Form fields have the `beforeInput` and `afterInput` slots
* GOV.UK service navigation added [`slots` for custom HTML](https://design-system.service.gov.uk/components/service-navigation/#use-slots-to-add-custom-html-code)

I think @paulrobertlloyd had ideas for this

anandamaryon1: A built-in approach would be nice, and my usage here maybe points to needing `start` and `end` for the visually hidden text, where currently the `badge` `visuallyHiddenText` supports only `start`/`prefix`.

colinrotherham: I've added support for visually hidden text placement in:

* https://github.com/nhsuk/nhsuk-frontend/pull/2099

```patch
  {{ badge({
    text: "New",
-   visuallyHiddenText: "feature"
+   visuallyHidden: {
+     text: "feature",
+     placement: "end"
+   }
  }) }}
```

PR #2559, opened 2026-06-01
Update radio and checkbox examples

anandamaryon1:
## Description
Swap out radios default example, to avoid using the same as inline example but not inline. 
And update guidance in line with gov's explanation, which I think is clearer. 

Also, add new example and guidance for inline checkboxes, introduced in frontend: https://github.com/nhsuk/nhsuk-frontend/commit/c7c5f1c2087291669b5bf7ea4b4ff1eb994175f1. 

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/views/design-system/components/checkboxes/inline/index.njk
+++ b/app/views/design-system/components/checkboxes/inline/index.njk
@@ -1,0 +1,24 @@
+{% from "checkboxes/macro.njk" import checkboxes %}
+
+{{ checkboxes({
+  idPrefix: "example-inline",
+  name: "exampleInline",
+  inline: true,
+  fieldset: {
+    legend: {
+      text: "Which nipple has changed?",
+      size: "l",
+      isPageHeading: true
+    }
+  },
+  items: [
+    {
+      value: "right",
+      text: "Right nipple"
+    },
+    {
+      value: "left",
+      text: "Left nipple"
+    }
+  ]
+}) }}
--- a/app/views/design-system/components/checkboxes/inline/index.njk
+++ b/app/views/design-system/components/checkboxes/inline/index.njk
@@ -3,7 +3,7 @@
 {{ checkboxes({
   idPrefix: "example-inline",
   name: "exampleInline",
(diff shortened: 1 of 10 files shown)


app/views/design-system/components/radios/default/index.njk
```diff
@@ -1,23 +1,32 @@
...
+      text: "Email"
+    },
+    {
+      value: "phone",
+      text: "Phone"
     },
     {
-      value: "no",
-      text: "No"
+      value: "text-message",
```
colinrotherham: Would you mind keeping the values as `email`, `phone` and `text`?

We use this example a lot and it keeps it consistent

```suggestion
      value: "text",
```

app/views/design-system/components/checkboxes/inline/index.njk
```diff
@@ -0,0 +1,24 @@
...
+    {
+      value: "right",
+      text: "Right nipple"
+    },
+    {
+      value: "left",
+      text: "Left nipple"
+    }
+  ]
+}) }}
```
colinrotherham: Can you check `.editorconfig` is working?

Just to catch these missing new lines at the end of files

app/views/design-system/components/checkboxes/inline/index.njk
```diff
@@ -0,0 +1,24 @@
+{% from "checkboxes/macro.njk" import checkboxes %}
+
+{{ checkboxes({
+  idPrefix: "example-inline",
+  name: "exampleInline",
+  inline: true,
```
colinrotherham: Did you want to use the class instead?

app/views/design-system/components/radios/default/index.njk
anandamaryon1: thanks, done

colinrotherham (approved):

sarawilcox (approved):
Looks good to me.

colinrotherham (approved):

PR #2571, opened 2026-06-10
Drop heading size on homepage to L (from XL)

frankieroberto:
Feels a bit more readable at large size rather than extra-large?

The large size is also used on comparable product pages:

https://prototype-kit.service-manual.nhs.uk
https://notify.nhs.uk
https://www.ravs.england.nhs.uk

## Screenshots

| Before | After |
| ------ | ----- |
| Screenshot 2026-06-10 at 11 05 28 | Screenshot 2026-06-10 at 11 05 07 |


anandamaryon1:
Thanks for this suggestion, I'm going to stick with the XL for now because of the knock-on effect of the other headings down the page and the fact we use XL across the rest of the website.

PR #2583, opened 2026-06-17
Add guidance for 'All' checkbox option

frankieroberto:
Also updates the existing guidance to use the renamed options.

See https://github.com/nhsuk/nhsuk-frontend/pull/1707 for the new component option.

## Screenshot

Screenshot 2026-06-17 at 16 23 57

### DIFF
--- a/app/views/design-system/components/checkboxes/index.njk
+++ b/app/views/design-system/components/checkboxes/index.njk
@@ -104,9 +104,9 @@
 
   <h3 id="all-option">Add an option for "all"</h3>
 
-  <p>You can give users an option to be able to quick select or unselect all the checkboxes. This checkbox is only shown when JavaScript is available.</p>
+  <p>You can give users an option to be able to quickly select or unselect all the checkboxes. This checkbox only displays when JavaScript is available.</p>
 
-  <p>This option is usually shown first, as otherwise users may not see it before starting to select each checkbox individually.</p>
+  <p>Show this option first so that users see it before starting to select each checkbox individually.</p>
 
   <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox container, and to any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
 
--- a/app/views/design-system/components/checkboxes/index.njk
+++ b/app/views/design-system/components/checkboxes/index.njk
@@ -108,7 +108,7 @@
 
   <p>Show this option first so that users see it before starting to select each checkbox individually.</p>
 
-  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox container, and to any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
+  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox, and add the HTML class `nhsuk-frontend-supported-only` to its container and any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
 
   {{ designExample({
     group: "components",
--- a/app/views/design-system/components/checkboxes/index.njk
+++ b/app/views/design-system/components/checkboxes/index.njk
@@ -66,7 +66,7 @@
     item: "checkboxes",
     type: "inline"
   }) }}
-  
+
   <h3 id="checkboxes-with-hint-text">Checkboxes with hint text</h3>
   <p>Unlike with <a href="/design-system/components/radios">radios</a>, users can select more than 1 option from a list of checkboxes. Do not assume that users will know how many options they can select.</p>
   <p>Use hint text to <a href="/content/how-to-write-good-questions-for-forms/write-the-supporting-content-for-your-form#give-users-help-in-context">give users help in context</a>. For example, say "Select all the options that are relevant to you". Read more about <a href="/design-system/components/hint-text">hint text</a>.</p>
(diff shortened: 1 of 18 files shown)


sarawilcox:
Hi @frankieroberto , I tweaked the content. Is this OK?

Screenshot 2026-06-18 at 12 18 11

sarawilcox (approved):
Looks good to me. I suggested minor tweak. Otherwise OK. :-)

anandamaryon1:
Do we have a standard way to mention code features? Here we mention the html and Nunjucks options in the text. But I have a feeling we sometimes mention only one or the other…

Here's an example from buttons:
image

Maybe something to follow, or discuss a new approach. It'll be more of a thing in v11 when we have more Nunjucks options for things that are just classes right now.

frankieroberto:
@anandamaryon1 I wasn’t sure how best to reference the Nunjucks and HTML options either. On the one hand we could not mention it at all and leave people to figure it out from the 2 code examples below. But I felt that may be a bit too hidden and hard to interpret?

app/views/design-system/components/checkboxes/index.njk
```diff
@@ -102,6 +102,20 @@
 
   <p>If JavaScript is unavailable, and a user selects both the "none" checkbox and another checkbox, display an error message.</p>
 
+  <h3 id="all-option">Add an option for "all"</h3>
+
+  <p>You can give users an option to be able to quickly select or unselect all the checkboxes. This checkbox only displays when JavaScript is available.</p>
+
+  <p>Show this option first so that users see it before starting to select each checkbox individually.</p>
+
+  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox container, and to any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
```
colinrotherham: This line is quite different to the instructions for `data-behaviour="exclusive"`

But also the container (and divider) needs `nhsuk-frontend-supported-only` not the data attribute

```suggestion
  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox, and add the HTML class `nhsuk-frontend-supported-only` to its container and any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
```

frankieroberto: Ah, well remembered!

colinrotherham: No problem! I've [updated the changelog entry](https://github.com/nhsuk/nhsuk-frontend/pull/1707/changes/141b18bc187df06965db4333bb8bc74758171bbb) to show this

Feedback welcome

colinrotherham:
Noticed we say "none of the above" and "select all" on NHS.UK frontend

Are these terms clearer than "none" and "all"?

(Same with inclusive and exclusive)

frankieroberto:
@colinrotherham I've tended to avoid "none of the above" even though that's how people sometimes refer to in out loud. "the above" implies a spatial framing which might make less sense to screen reader users?

Often in context I'd recommend using "none of these" or something more specific like "none of these countries".

colinrotherham:
Mainly asking so we can update the NHS.UK frontend examples to "none" and "all" so they match

sarawilcox:
Hi Frankie, we mention a class here: 

Screenshot 2026-06-22 at 11 09 36

Elsewhere in the design system we present these as code (in red). Should this be the same?

Screenshot 2026-06-22 at 11 08 46

Screenshot 2026-06-22 at 11 09 15

sarawilcox:
I agree that we should avoid "none of the above" or "all of the below". 

In the "none" option for checkboxes, we say: "No, I do not have any of these symptoms " which is rather wordy. 
https://service-manual.nhs.uk/design-example/components/checkboxes/none-option 

In the FE, we say: "None of the above "
https://nhsuk.github.io/nhsuk-frontend/components/checkboxes/with-none-of-the-above-option/

Have you tested "All" on its own, Frankie?

app/views/design-system/components/checkboxes/index.njk
```diff
@@ -102,6 +102,20 @@
 
   <p>If JavaScript is unavailable, and a user selects both the "none" checkbox and another checkbox, display an error message.</p>
 
+  <h3 id="all-option">Add an option for "all"</h3>
+
+  <p>You can give users an option to be able to quickly select or unselect all the checkboxes. This checkbox only displays when JavaScript is available.</p>
+
+  <p>Show this option first so that users see it before starting to select each checkbox individually.</p>
+
+  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox, and add the HTML class `nhsuk-frontend-supported-only` to its container and any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
```
frankieroberto:
```suggestion
  <p>To use it, add the <code>data-behaviour="inclusive"</code> attribute to the checkbox, and add the HTML class <code>nhsuk-frontend-supported-only</code> to its container and any divider that appears after it. If you are using the Nunjucks macro, use the <code>behaviour: "inclusive"</code> option instead.</p>
```

frankieroberto:
> Hi Frankie, we mention a class here:
> 
> Screenshot 2026-06-22 at 11 09 36

Opps, this was my mistake. I used the markdown for the code reference (backticks) rather than the HTML `<code>`. Fixed now - thanks for spotting!

frankieroberto:
> I agree that we should avoid "none of the above" or "all of the below".
> 
> In the "none" option for checkboxes, we say: "No, I do not have any of these symptoms " which is rather wordy. https://service-manual.nhs.uk/design-example/components/checkboxes/none-option
> 
> In the FE, we say: "None of the above " https://nhsuk.github.io/nhsuk-frontend/components/checkboxes/with-none-of-the-above-option/

I think the longer more wordy option might be better here, just to help prevent mistakes where a user misreads or skips the question. I often end up doing that for Yes/No questions too, where the answer partially repeats the question.

> Have you tested "All" on its own, Frankie?

No - I think it should probably be "All vaccines" or perhaps "All 9 vaccines" in the example? (sometimes the number can help).

The example we tested in RAVS used "All 21 organisations" as the label: https://github.com/nhsuk/nhsuk-frontend/pull/1707#issuecomment-3632066144

sarawilcox:
I like "All 9 vaccines". It's specific.

anandamaryon1:
Pushed an update to use "All 9 vaccines". I'll go and update the example on frontend to macth

anandamaryon1 (approved):

PR #1296, opened 2021-07-08
Update 9-respect-and-protect-users-confidentiality-and-privacy.njk

sarawilcox:
Correct apostrophe typo

PR #2585, opened 2026-06-19
Add a reference page for override classes

frankieroberto:
The override classes are (sensibly) referenced in context in the relevant pages.

However designers have asked if there can also be a single reference page listing them all. 

This adds that. It also lets us describe some override classes which are currently not mentioned in the rest of the documentation at all.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/views/design-system/guides/override-classes.njk
+++ b/app/views/design-system/guides/override-classes.njk
@@ -12,7 +12,7 @@
 {% endblock %}
 
 {% block bodyContent %}
-  <p>You can change the default appearance of components by using override classes.</p>
+  <p>You can change the default appearance of components by using override classes, sometimes known as utility classes.</p>
 
   <p>Override classes provided by the design system all have the prefix <code>nhsuk-u-</code>.</p>
 
--- a/app/views/design-system/guides/override-classes.njk
+++ b/app/views/design-system/guides/override-classes.njk
@@ -1,0 +1,418 @@
+{% extends "layouts/app-layout.njk" %}
+
+{% set pageTitle = "Override classes" %}
+{% set pageDescription = "A reference list of all override classes provided by the design system, grouped by category." %}
+{% set pageSection = "Design system" %}
+{% set subSection = "Design system" %}
+{% set theme = "Guides" %}
+{% set dateUpdated = "August 2025" %}
+
+{% block beforeContent %}
+  {% include "design-system/_breadcrumb.njk" %}
+{% endblock %}
+
+{% block bodyContent %}
+  <p>You can change the default appearance of components by using override classes.</p>
+
+  <p>Override classes provided by the design system all have the prefix <code>nhsuk-u-</code>.</p>
+
+  <h2>Spacing</h2>
+
+  <p>Spacing override classes control the margin and padding on elements. They come in 2 types: responsive and static. See the <a href="/design-system/styles/spacing">spacing style page</a> for guidance on when and how to use spacing.</p>
+
+  <h3>Responsive spacing</h3>
+
+  <p>Responsive spacing classes apply different values on mobile and on tablet and larger screens. You can replace <code>0</code> with any number from 0 to 9.</p>
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-margin-0</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "Responsive margin on all sides" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-margin-top-0</code>" },
+        value: { text: "Responsive margin on the top" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-margin-right-0</code>" },
+        value: { text: "Responsive margin on the right" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-margin-bottom-0</code>" },
+        value: { text: "Responsive margin on the bottom" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-margin-left-0</code>" },
+        value: { text: "Responsive margin on the left" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-padding-0</code>" },
+        value: { text: "Responsive padding on all sides" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-padding-top-0</code>" },
+        value: { text: "Responsive padding on the top" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-padding-right-0</code>" },
+        value: { text: "Responsive padding on the right" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-padding-bottom-0</code>" },
+        value: { text: "Responsive padding on the bottom" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-padding-left-0</code>" },
+        value: { text: "Responsive padding on the left" }
+      }
+    ]
+  }) }}
+
+  <h3>Static spacing</h3>
+
+  <p>Static spacing classes apply a fixed value at all screen sizes. You can replace <code>0</code> with any number from 0 to 9.</p>
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-static-margin-0</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "Fixed margin on all sides" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-margin-top-0</code>" },
+        value: { text: "Fixed margin on the top" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-margin-right-0</code>" },
+        value: { text: "Fixed margin on the right" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-margin-bottom-0</code>" },
+        value: { text: "Fixed margin on the bottom" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-margin-left-0</code>" },
+        value: { text: "Fixed margin on the left" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-padding-0</code>" },
+        value: { text: "Fixed padding on all sides" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-padding-top-0</code>" },
+        value: { text: "Fixed padding on the top" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-padding-right-0</code>" },
+        value: { text: "Fixed padding on the right" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-padding-bottom-0</code>" },
+        value: { text: "Fixed padding on the bottom" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-static-padding-left-0</code>" },
+        value: { text: "Fixed padding on the left" }
+      }
+    ]
+  }) }}
+
+  <h2>Layout</h2>
+
+  <p>See the <a href="/design-system/styles/layout">layout style page</a> for guidance on using the grid system.
+
+  <h3>Grid column widths</h3>
+
+  <p>Use these classes to set the width of grid columns. They apply <code>float: left</code> alongside the width. Use the <code>-tablet</code> variants if you want the column to expand to full width on mobile.</p>
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-one-half</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "50% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-third</code>" },
+        value: { text: "33.33% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-two-thirds</code>" },
+        value: { text: "66.66% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-quarter</code>" },
+        value: { text: "25% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-three-quarters</code>" },
+        value: { text: "75% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-half-tablet</code>" },
+        value: { text: "50% width on tablet and above, 100% on mobile" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-third-tablet</code>" },
+        value: { text: "33.33% width on tablet and above, 100% on mobile" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-two-thirds-tablet</code>" },
+        value: { text: "66.66% width on tablet and above, 100% on mobile" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-quarter-tablet</code>" },
+        value: { text: "25% width on tablet and above, 100% on mobile" }
+      },
+      {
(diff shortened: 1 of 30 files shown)


Demotive:
FWIW I commend this page - I knew a few of these and have found them very useful. Now I know more, and will find them useful. Good resource to have published I'd say

anandamaryon1:
Nice, thanks for this @frankieroberto! Very helpful. 

Override is probably a better name than utility but, now begs the question of why they're `nhsuk-u` instead of `nhsuk-o` (not suggesting we change the CSS). Maybe we can just mention that they're sometimes referred to as utility classes too?

frankieroberto:
@anandamaryon1 I'm not actually sure why the NHS design system originally picked the `nhsuk-u-` prefix instead of GOVUK's `govuk-!-`!

Good idea to also mention "utility" on the page. I used to call them "modifier classes" for some reason!

app/views/design-system/guides/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+{% set subSection = "Design system" %}
+{% set theme = "Guides" %}
+{% set dateUpdated = "August 2025" %}
+
+{% block beforeContent %}
+  {% include "design-system/_breadcrumb.njk" %}
+{% endblock %}
+
+{% block bodyContent %}
+  <p>You can change the default appearance of components by using override classes.</p>
```
frankieroberto:
```suggestion
  <p>You can change the default appearance of components by using override classes, sometimes known as utility classes.</p>
```

sarawilcox:
Hi @frankieroberto , sounds like a good idea. :heart: 

I was a bit surprised about its location. I expected it in the styles section rather than in a separate guide. 

We've got some content about font override classes in the typography section: https://service-manual.nhs.uk/design-system/styles/typography#font-override-classes. And width override classes in layout: https://service-manual.nhs.uk/design-system/styles/layout#width-override-classes. We should link these 2 sections to the fuller page. But perhaps also put them in the same place? (I'm not a designer though, so not a user who'd be looking for override classes.)

There's also the spacing section: https://service-manual.nhs.uk/design-system/styles/spacing. I'm now wondering if there's too much duplication across these various pages? Could we/should we cut back the detail on some of the other pages if we have the full list in 1 place? 

The other 2 pages in the Guides section are about updating to v10 or WCAG 2.2. I feel that override classes doesn't sit well next to them. 

Also, elsewhere, we code the classes up in red. 
Screenshot 2026-06-23 at 13 27 14

frankieroberto:
@sarawilcox I wasn't sure where it should go - styles section might be better!

I think the duplication with other pages is ok. In my head this page is a pure reference, mainly for designers who want to quickly remind themselves what the options are. And then they also get explained in context within relevant pages.

The reference guide links to the pages you mentioned already, plus the [Color](https://nhsuk-service-manual-pr-2585.herokuapp.com/design-system/styles/colour) page which mentions the secondary text colour class.

The other page it links to is the [fluid width classes](https://nhsuk-service-manual-pr-2585.herokuapp.com/design-system/components/text-input#fluid-width-inputs) section of the Text input page - this is within Components, rather than Styles - which might be one reason to put this in Guides rather than Styles?

> Also, elsewhere, we code the classes up in red.

I noticed this too. I think the CSS needs updating so that `<code>` still gets the red-on-white styling even when it's in a table header cell. Will investigate!

sarawilcox:
If other people think it's OK in the guides section, I'm happy with that. We can link to it from the other pages.

anandamaryon1:
I also looked in Styles first, but then when I saw it in Guides thought, ok that makes sense. 

…But, given we both looked in styles first maybe it should be there? It is more of a reference than a guide too.

frankieroberto:
Happy to move it to Styles, but where would we put it in the nav? Under a new heading?

anandamaryon1:
Let's try it as it's own section under Styles @frankieroberto

frankieroberto:
> Let's try it as it's own section under Styles @frankieroberto

@anandamaryon1 Done. Have listed it under "Reference" in the styles section. That seem ok?

anandamaryon1:
I like it, thank you

anandamaryon1:
Just need a content review from @sarawilcox and we can include in next release

anandamaryon1:
Odd side nav bug, the "References" section seems to disappear when navigating to any of the other "Styles" sections that have nested items, or the page: "[Use the NHS Frutiger font](https://nhsuk-service-manual-pr-2585.herokuapp.com/design-system/styles/use-frutiger-font)"

frankieroberto:
> Odd side nav bug, the "References" section seems to disappear when navigating to any of the other "Styles" sections that have nested items, or the page: "[Use the NHS Frutiger font](https://nhsuk-service-manual-pr-2585.herokuapp.com/design-system/styles/use-frutiger-font)"

Good spot! Seems those pages have a semi-duplicated side nav. Fixed in 5334d66f3baec4145af0de834693f4aa58ea18f9

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+{% endblock %}
+
+{% block bodyContent %}
+  <p>You can change the default appearance of components by using override classes, sometimes known as "utility classes".</p>
+
+  <p>Override classes provided by the design system all have the prefix <code>nhsuk-u-</code>.</p>
+
+  <h2>Spacing</h2>
+
+  <p>Spacing override classes control the margin and padding on elements. They come in 2 types: responsive and static. See the <a href="/design-system/styles/spacing">spacing style page</a> for guidance on when and how to use spacing.</p>
```
colinrotherham: This page is looking ace @frankieroberto

Slight nitpick, but it's the "Styles" section (plural) in both the URL and side nav

```suggestion
  <p>Spacing override classes control the margin and padding on elements. They come in 2 types: responsive and static. See the <a href="/design-system/styles/spacing">spacing styles page</a> for guidance on when and how to use spacing.</p>
```

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-one-half</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "50% width at all screen sizes" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-one-third</code>" },
+        value: { text: "33.33% width at all screen sizes" }
```
colinrotherham: Can we spell these out?

Mainly because it's a daft precision at ~`33.33%`~ `33.3333333333%`

But from our content guide:

>### [Fractions and percentages](https://service-manual.nhs.uk/content/numbers-measurements-dates-time#fractions-and-percentages)
>
>Spell out common fractions like two-thirds.
>
>For percentages, use the % symbol, for example, 50%. However, it's often better not to use a percentage. Instead of 50%, for example, you could say "1 in 2" or "half".

sarawilcox: Well spotted. Thanks Colin. @frankieroberto I've made the changes. Would one of you mind checking that I've got them right, please?

app/views/design-system/styles/override-classes.njk
sarawilcox: Done.

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+
+  <h3>Responsive spacing</h3>
+
+  <p>Responsive spacing classes apply different values on mobile and on tablet and larger screens. You can replace <code>0</code> with any number from 0 to 9.</p>
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-margin-0</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "Responsive margin on all sides" }
```
colinrotherham, line 31: I know this is for responsive margins, but zero is never responsive

Shall we update all these descriptions accordingly?

```suggestion
        value: { text: "Remove margin on all sides" }
```

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+      },
+      {
+        key: { html: "<code>nhsuk-u-margin-top-0</code>" },
+        value: { text: "Responsive margin on the top" }
```
colinrotherham, line 35:
```suggestion
        value: { text: "Remove margin on the top" }
```

app/views/design-system/styles/override-classes.njk
colinrotherham: Ah just seen we use `0` as the placeholder example!

In that case, can we use a better non-zero example?

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+    ]
+  }) }}
+
+  <h2>Layout</h2>
+
+  <p>See the <a href="/design-system/styles/layout">layout styles page</a> for guidance on using the grid system.
+
+  <h3>Grid column widths</h3>
+
+  <p>Use these classes to set the width of grid columns. They apply <code>float: left</code> alongside the width. Use the <code>-tablet</code> variants if you want the column to expand to full width on mobile.</p>
```
colinrotherham: These have been removed in v11 and shouldn't be documented

Can we document the `nhsuk-grid-column-*` classes instead? 🙏

@anandamaryon1 Maybe we should announce them as deprecated in v10.x? Although I don't think we did because grid classes are applied at desktop and these ones at tablet

Here are the changelog entries from v11 when they become redundant (both from tablet)

<br>

> ### Mobile width override classes
> 
> Grid column classes ending `-from-mobile` are applied at mobile width (320px) and above.
> 
> If you're using the following mobile width utility classes, you must:
> 
> - Replace `nhsuk-u-one-half` with `nhsuk-grid-column-one-half-from-mobile`
> - Replace `nhsuk-u-one-third` with `nhsuk-grid-column-one-third-from-mobile`
> - Replace `nhsuk-u-two-thirds` with `nhsuk-grid-column-two-thirds-from-mobile`
> - Replace `nhsuk-u-one-quarter` with `nhsuk-grid-column-one-quarter-from-mobile`
> - Replace `nhsuk-u-three-quarters` with `nhsuk-grid-column-three-quarters-from-mobile`.
>
> ### Tablet width override classes
> 
> Grid column classes not ending `-from-mobile` or `-from-desktop` are applied at tablet width (641px) and above.
> 
> If you're using the following tablet width utility classes, you must:
> 
> - Replace `nhsuk-u-one-half-tablet` with `nhsuk-grid-column-one-half`
> - Replace `nhsuk-u-one-third-tablet` with `nhsuk-grid-column-one-third`
> - Replace `nhsuk-u-two-thirds-tablet` with `nhsuk-grid-column-two-thirds`
> - Replace `nhsuk-u-one-quarter-tablet` with `nhsuk-grid-column-one-quarter`
> - Replace `nhsuk-u-three-quarters-tablet` with `nhsuk-grid-column-three-quarters`
> 
> This change was introduced in [pull request #1296: Apply grid classes from tablet (not desktop)](https://github.com/nhsuk/nhsuk-frontend/pull/1296).

app/views/design-system/styles/override-classes.njk
frankieroberto: I wasn't sure how best to show that the number is variable. Using `n` or `{n}` is one option, but that could be a bit confusing and geeky, so I went with zero.

Could swap out to `1` or `2` though? What do you think @sarawilcox and @anandamaryon1?

app/views/design-system/styles/override-classes.njk
frankieroberto: Resolving this, as we can discuss an approach in https://github.com/nhsuk/nhsuk-service-manual/pull/2585#discussion_r3498904524 and then apply to both.

app/views/design-system/styles/override-classes.njk
frankieroberto: @colinrotherham if this page goes live soon, with the next v10 release, then I think it's worth documenting _all_ overrides as they exist now, just to establish the page as being comprehensive?

But we could add a deprecation note at the top if they're going to be deprecated?

app/views/design-system/styles/override-classes.njk
frankieroberto: Happy to spell these out.

We should also update https://service-manual.nhs.uk/design-system/styles/layout#width-override-classes though 😊  – can do that in a separate PR?

anandamaryon1: > Happy to spell these out.
> 
> We should also update https://service-manual.nhs.uk/design-system/styles/layout#width-override-classes though 😊 – can do that in a separate PR?

Shall we just do those here? @sarawilcox

app/views/design-system/styles/override-classes.njk
anandamaryon1: Let's go for 3 since it changes based on screen size

app/views/design-system/styles/override-classes.njk
sarawilcox: Yes, I think let's just do them. I can do it tomorrow, unless anyone wants to do them today.

app/views/design-system/styles/override-classes.njk
```diff
@@ -0,0 +1,418 @@
...
+
+  {{ summaryList({
+    rows: [
+      {
+        key: { html: "<code>nhsuk-u-text-break-word</code>", classes: "nhsuk-u-width-one-half" },
+        value: { text: "Forces long words or URLs to break onto a new line" }
+      },
+      {
+        key: { html: "<code>nhsuk-u-nowrap</code>" },
+        value: { text: "Prevents a link from wrapping onto multiple lines on mobile" }
```
edwardhorsford: Does this class only work for links/anchors? if not, make example less specific?

edwardhorsford: Note this class is largely useless until v11

app/views/design-system/styles/override-classes.njk
colinrotherham: Go for it if @anandamaryon1 is happy

Otherwise consider them a mistake, we shouldn't have had two sets of column classes

app/views/design-system/styles/override-classes.njk
anandamaryon1:
```suggestion
        value: { text: "Prevents text from wrapping onto multiple lines on mobile" }
```
Text?

app/views/design-system/styles/override-classes.njk
anandamaryon1: Yeah makes sense to have it accurate right now and we'll need to update for v11

app/views/design-system/styles/override-classes.njk
frankieroberto: @edwardhorsford why is it largely useless until v11?

anandamaryon1: Because it's on mobile only

frankieroberto: Ah! And we’re extending it to all sizes in v11? I missed that, sorry. Seems sensible though!

app/views/design-system/styles/override-classes.njk
colinrotherham: Can we definitely mention that they're deprecated then, and refer people to the [grid system](https://service-manual.nhs.uk/design-system/styles/layout#grid)?

app/views/design-system/styles/override-classes.njk
sarawilcox: I've fixed them.

anandamaryon1 (approved):
Thanks for this @frankieroberto it's a great addition

PR #2647, opened 2026-08-28
Add documentation for additional page template variables

frankieroberto:
These were added in https://github.com/nhsuk/nhsuk-frontend/pull/1795 but we forgot to add them to the documentation.

### DIFF
--- a/app/views/design-system/styles/page-template/index.njk
+++ b/app/views/design-system/styles/page-template/index.njk
@@ -294,7 +294,7 @@
           text: "Variable"
         },
         {
-          html: "Add a nonce attribute to the script for your Content Security Policy (CSP). Provide a nonce that hostile actors cannot guess, as otherwise they can easily find a way around your CSP. However, you should use this attribute only if you’re not able to include the hash for the inline scripts in your service's CSP."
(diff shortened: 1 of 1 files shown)


colinrotherham (approved):

app/views/design-system/styles/page-template/index.njk
```diff
@@ -286,6 +286,17 @@
...
       ],
+      [
+        {
+          text: "cspNonce"
+        },
+        {
+          text: "Variable"
+        },
+        {
+          html: "Add a nonce attribute to the script for your Content Security Policy (CSP). Provide a nonce that hostile actors cannot guess, as otherwise they can easily find a way around your CSP. However, you should use this attribute only if you’re not able to include the hash for the inline scripts in your service's CSP."
```
colinrotherham: Thanks for adding these 🙌

Sorry just realised we've got mixed curly/straight apostrophes

```suggestion
          html: "Add a nonce attribute to the script for your Content Security Policy (CSP). Provide a nonce that hostile actors cannot guess, as otherwise they can easily find a way around your CSP. However, you should use this attribute only if you're not able to include the hash for the inline scripts in your service's CSP."
```

frankieroberto: @colinrotherham thanks - thought I'd double-checked this but I missed one!

PR #2662, opened 2026-09-09
Add hidden text to warning button

sarawilcox:
## Description

There's currently nothing in this button to indicate to screen reader users that it's red to act as a warning. Add hidden text.

@colinrotherham, have I coded this correctly, please?

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/views/design-system/components/buttons/warning/index.njk
+++ b/app/views/design-system/components/buttons/warning/index.njk
@@ -1,6 +1,6 @@
 {% from "button/macro.njk" import button %}
 
 {{ button({
-  text: "Yes, delete this vaccine",
+  html: '<span class="nhsuk-u-visually-hidden">Warning: </span>Yes, delete this vaccine',
   variant: "warning"
 }) }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS digital service manual Changelog
 
+## TBC
+
+:wrench: **Maintenance**
+
+- Add hidden text to red warning button for screen reader users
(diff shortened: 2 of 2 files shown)


colinrotherham (approved):

app/views/design-system/components/buttons/warning/index.njk
```diff
@@ -1,6 +1,6 @@
 {% from "button/macro.njk" import button %}
 
 {{ button({
-  text: "Yes, delete this vaccine",
+  html: '<span class="nhsuk-u-visually-hidden">Warning: </span>Yes, delete this vaccine',
```
colinrotherham, line 4: @anandamaryon1 Will we want to automatically add this to `variant: "warning"`?

And if should, should it be customisable, e.g. `visuallyHiddenText: "Urgent"`

But then, [based on our API proposal](https://github.com/nhsuk/nhsuk-frontend/pull/2099#issuecomment-5779078104), what if we want visually hidden text before _and_ after?

anandamaryon1: Yeah I was imagining we add it by default, like we do with eg. Warning callout's heading. 

Yes the text should be customisable. 

Not sure on the need for after. Is the issue that the `visuallyHiddenText` will consume a `slot`?

colinrotherham: Not that it consumes a slot, but just in case a "Delete" button has visually hidden text making it "Delete Colin Rotherham" 😄

Bit like we do for change links

